PR #274, opened 2022-07-15
feat: add hook for responsive blocks

SamuelAlev:


SamuelAlev:
Also see Tailwind Container Queries official plugin

PR #280, opened 2022-07-18
chore(shared): preserve module structure when building

triggertoo:


packages/shared/tsconfig.json
```diff
@@ -14,5 +14,9 @@
         "outDir": "dist"
     },
     "include": ["src/**/*"],
-    "exclude": ["node_modules", "dist"]
+    "exclude": [
+        "dist",
+        "node_modules",
+        "src/**/*.spec.ts",
```
SamuelAlev, line 20: We need the tests to access the tsconfig, why excluding?

triggertoo: by mistake, removed it

rawi96 (approved):

Walkeezy (approved):

PR #281, opened 2022-07-19
feat(PersonalNoteBlock): add saving of plaintext content

Walkeezy:
This PR adds saving of the blocks content as plaintext to the settings.


packages/personal-note-block/src/PersonalNoteBlock.tsx
```diff
@@ -66,6 +66,7 @@ export const PersonalNoteBlock: FC<BlockProps> = ({ appBridge }) => {
         setBlockSettings({
             ...blockSettings,
             note: value,
```
SamuelAlev, line 68: Please don’t save the data twice

PR #288, opened 2022-07-28
chore: update app bridge to v3.0.0-beta.5

triggertoo:


triggertoo:
seems like we need to update cypress in terms of `structuredClone is not defined`

PR #294, opened 2022-08-05
Figma block with possibility to toggle between svg image and svg with additional functionality

bojangles-m:


packages/figma-block/src/settings.ts
```diff
@@ -71,12 +72,20 @@ export const settings: BlockSettings = {
         },
     ],
     layout: [
+        {
+            id: HAS_FUNCTIONALITY_ID,
+            type: 'switch',
+            label: 'Functionality',
+            defaultValue: false,
+            info: 'This is much more fun :)',
```
SamuelAlev, line 80: Is the tooltip text like that in the design? 😂

bojangles-m: You are to fast this time :), will let you know when it is ready, I am waiting for this :) but thnx for checking the PR

SamuelAlev (approved):

PR #302, opened 2022-08-10
feat(Storybook Block):Multiple minor fixes

rawi96:
ticket: https://app.clickup.com/t/2t7u1we

### DIFF
--- a/packages/storybook-block/src/types.ts
+++ b/packages/storybook-block/src/types.ts
@@ -25,7 +25,7 @@ export type Settings = {
 
 export enum StorybookStyle {
     Default = 'Default',
-    WithoutAddons = 'WithoutAddons',
+    WithAddons = 'WithAddons',
 }
 
 export enum StorybookBorderRadius {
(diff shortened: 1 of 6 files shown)


packages/storybook-block/src/types.ts
```diff
@@ -58,11 +58,11 @@ export const borderStyles: Record<StorybookBorderStyle, string> = {
...
-export const borderRadiusClasses: Record<StorybookBorderRadius, string> = {
-    [StorybookBorderRadius.None]: 'tw-rounded-none',
-    [StorybookBorderRadius.Small]: 'tw-rounded',
-    [StorybookBorderRadius.Medium]: 'tw-rounded-md',
-    [StorybookBorderRadius.Large]: 'tw-rounded-lg',
+export const borderRadiuses: Record<StorybookBorderRadius, string> = {
+    [StorybookBorderRadius.None]: '0px',
+    [StorybookBorderRadius.Small]: '0.25rem',
+    [StorybookBorderRadius.Medium]: '0.375rem',
+    [StorybookBorderRadius.Large]: '0.5rem',
```
Walkeezy, line 65: Where do these values come from? We have a `radiusStyleMap` with different values [in the shared package](https://github.com/Frontify/guideline-blocks/blob/5541381e3f11617502462f14b7e584a21c01b813/packages/shared/src/settings/types.ts#L24) (some other blocks are using the values from the shared package). Should we maybe adjust these values there too?

rawi96: I just took the values that were behind the tailwind classes.
But the radiusStyleMap makes more sense. I changed it and now the radiusStyleMap is used. 😊

celine-s (approved):
nice 👍

Walkeezy (approved):

PR #306, opened 2022-08-17
feat: change Color utilities to new color format

Walkeezy:
Adapt shared color utilites to handle the new long notation color format.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
         "vitest": "0.20.3"
     },
     "dependencies": {
-        "@frontify/fondue": "9.11.0",
+        "@frontify/fondue": "11.0.0",
         "react": "17.0.2",
         "react-dom": "17.0.2"
     }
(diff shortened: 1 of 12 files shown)


bojangles-m (approved):
For Figma block looks ok.

tsconfig.json
```diff
@@ -13,5 +13,5 @@
         "resolveJsonModule": true,
         "types": ["cypress", "cypress-real-events"]
     },
-    "exclude": ["**/dist", "**/node_modules"]
+    "exclude": ["**/dist"]
```
rawi96, line 16: please remove this change.

Walkeezy: Done

rawi96:
alright, changes look now good.

But what I realized:
When I save the long color format directly in the DB it just renders always black instead of the correct color.

Imho this should not be the case now anymore. 🤔

Walkeezy:
> But what I realized: When I save the long color format directly in the DB it just renders always black instead of the correct color.
Can't reproduce this. When I serve a storybook block from this PR on a local development block it works fine, when I change the color format to the long notation in the database.

celine-s (approved):
I could neither reproduce the error and changes look good to me 👍

PR #316, opened 2022-08-30
feat: add `debug` type

SamuelAlev:
Add `debug` setting type, it will allow to edit the dev to get a bit more control

Linked PR https://github.com/Frontify/clarify/pull/10710

### DIFF
--- a/packages/block-settings/types/blocks/debug.ts
+++ b/packages/block-settings/types/blocks/debug.ts
@@ -2,4 +2,5 @@
 
 export type DebugBlock = {
     type: 'debug';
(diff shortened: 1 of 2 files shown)


packages/block-settings/types/blocks/debug.ts
```diff
@@ -0,0 +1,6 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+export type DebugBlock = {
```
triggertoo, line 3: please extend `BaseBlock`, because value is now required

PR #317, opened 2022-08-30
fix(StorybookBlock): handle http:// and therefore localhost urls

celine-s:


packages/storybook-block/src/utils/buildIframeUrl.spec.ts
```diff
@@ -34,7 +34,7 @@ describe('buildIframeUrl', () => {
             StorybookPosition.Horizontal
         );
         expect(result?.toString()).toEqual(
-            'https://fondue-components.frontify.com/iframe.html?path=%2Fstory%2Fcomponents-action-menu--action-menu&nav=false&panel=false'
+            'http://fondue-components.frontify.com/iframe.html?path=%2Fstory%2Fcomponents-action-menu--action-menu&nav=false&panel=false'
```
rawi96, line 37: I would remove this test since it doesn't make any sense anymore.
_It should set an http url to https_ --> it does not change the protocol anymore

rawi96 (commented):
Just remove the test that I mentioned. The rest looks good.

rawi96 (approved):

packages/storybook-block/src/utils/ensureUrl.ts
```diff
@@ -0,0 +1,10 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+export const ensureUrl = (testUrl: string) => {
```
Walkeezy, line 3: As all this does is to check if a url already has a protocol (http or https) you could rename this to `addMissingUrlProtocol`, wdyt?

Walkeezy (approved):

packages/storybook-block/src/utils/ensureUrl.ts
celine-s: good point - thanks 👍

PR #323, opened 2022-09-05
feat(storybookBlock): add option to adjust height by dragging

celine-s:


packages/storybook-block/src/components/Resizable.tsx
```diff
@@ -0,0 +1,67 @@
+import { FC, useEffect, useState } from 'react';
```
Walkeezy, line 1:
```suggestion
/* (c) Copyright Frontify Ltd., all rights reserved. */

import { FC, useEffect, useState } from 'react';
```

packages/storybook-block/src/components/Resizable.tsx
```diff
@@ -0,0 +1,67 @@
...
+                {active && (
+                    <div className="tw-fixed tw-top-0 tw-bottom-0 tw-right-0 tw-left-0 tw-select-none tw-z-40" />
+                )}
+                {children}
+            </div>
+            <button
+                id="draghandle"
+                type="button"
+                onMouseDown={handler}
+                className="tw-cursor-ns-resize tw-w-full tw-pt-7 tw-h-full tw-bg-white tw-z-50"
```
Walkeezy, line 60: I would maybe move the drag handle closer to the element, so that it's clear to which element it belongs to. We have a little difference to the design, as we are missing the text below the iframe in the actual block for now ("only visible to people invited to this brand"). So I guess we could minimize the spacing there.

```suggestion
                className="tw-cursor-ns-resize tw-w-full tw-pt-3 tw-h-full tw-bg-white tw-z-50"
```

Walkeezy (approved):

rawi96 (approved):

PR #326, opened 2022-09-07
feat: simplify & fix Divider Block

Walkeezy:
Ticket: https://app.clickup.com/t/2t7u1ww

### DIFF
--- a/packages/divider-block/src/settings.ts
+++ b/packages/divider-block/src/settings.ts
@@ -89,8 +89,8 @@ export const settings: BlockSettings = {
                         {
                             id: WIDTH_CUSTOM_ID,
                             type: 'input',
-                            placeholder: '75%',
-                            clearable: false,
+                            placeholder: 'e.g. 60%',
+                            clearable: true,
                             rules: [numericalOrPercentRule, betweenPercentRule(0, 100)],
                             onChange: (bundle: Bundle): void => appendUnit(bundle, WIDTH_CUSTOM_ID, '%'),
                         },
@@ -163,8 +163,8 @@ export const settings: BlockSettings = {
                         {
                             id: HEIGHT_CUSTOM_ID,
                             type: 'input',
-                            placeholder: '100px',
-                            clearable: false,
+                            placeholder: 'e.g. 50px',
+                            clearable: true,
                             rules: [numericalOrPixelRule, minimumNumericalOrPixelOrAutoRule(10)],
                             onChange: (bundle: Bundle): void => appendUnit(bundle, HEIGHT_CUSTOM_ID),
                         },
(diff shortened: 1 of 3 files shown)


packages/divider-block/src/settings.ts
```diff
@@ -89,8 +77,8 @@ export const settings: BlockSettings = {
                         {
```
celine-s, line 77: I would remove the surrounding `sectionHeading`. Imo it is clear that it is a `Line` when Line is selected and other blocks don't have it either.

packages/divider-block/src/settings.ts
```diff
@@ -163,8 +151,8 @@ export const settings: BlockSettings = {
...
                             id: HEIGHT_CUSTOM_ID,
```
celine-s, line 152: here as well. the `sectionHeading` with label `Block` is imo not necessary. Otherwise I would remove the `Block` within the label `Block Height` (line 144).

packages/divider-block/src/settings.ts
```diff
@@ -202,70 +190,51 @@ export const settings: BlockSettings = {
...
-                        },
-                    ],
-                    off: [
-                        {
-                            id: THICKNESS_SIMPLE_ID,
-                            type: 'slider',
-                            defaultValue: THICKNESS_DEFAULT_VALUE,
+                            id: 'style',
+                            type: 'dropdown',
+                            label: 'Styling',
```
celine-s, line 204: maybe then add here `Line Styling` to have the same lable texts

packages/divider-block/src/settings.ts
```diff
@@ -202,70 +190,51 @@ export const settings: BlockSettings = {
...
-                                    label: 'L',
+                                    value: DividerStyle.Dashed,
+                                    label: 'Dashes',
                                 },
                             ],
                         },
+                        {
+                            id: THICKNESS_ID,
+                            type: 'input',
+                            label: 'Thickness',
```
celine-s, line 224: I would remove this label

packages/divider-block/src/settings.ts
```diff
@@ -202,70 +190,51 @@ export const settings: BlockSettings = {
...
+                            defaultValue: '1px',
+                            placeholder: 'e.g. 3px',
+                            clearable: false,
+                            rules: [numericalOrPixelRule, minimumNumericalOrPixelOrAutoRule(1)],
+                        },
+                        {
+                            id: 'color',
+                            type: 'colorInput',
+                            label: 'Color',
```
celine-s, line 233: And this label 🙃

celine-s (approved):
just some small comments about the labeling we had already before :D. You can also leave them like they are - I'm fine with either way.

packages/divider-block/src/settings.ts
celine-s: design is like this. Therefore, just keep it as it is

packages/divider-block/src/settings.ts
Walkeezy: Done

packages/divider-block/src/settings.ts
Walkeezy: Done

rawi96 (approved):

PR #327, opened 2022-09-07
feat(BlockSettings): add new type Setting

Walkeezy:
This PR adds a new type `Setting` to allow arrow functions on the `label`, `info` and `defaultValue` properties.

### DIFF
--- a/packages/block-settings/types/blocks/base.ts
+++ b/packages/block-settings/types/blocks/base.ts
@@ -2,12 +2,14 @@
 
 import { Bundle } from '../bundle';
 
+type Setting<T> = T | ((bundle: Bundle) => T);
+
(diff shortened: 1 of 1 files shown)


rawi96 (approved):

packages/block-settings/types/blocks/base.ts
```diff
@@ -2,12 +2,14 @@
...
 import { Bundle } from '../bundle';
 
+type Setting<T> = T | ((bundle: Bundle) => T);
+
 export type BaseBlock<T = undefined> = {
     id: string;
-    label?: string;
-    info?: string;
+    label?: Setting<string>;
+    info?: Setting<string>;
```
SamuelAlev, line 10: Can we use another name than `Setting`?

PR #340, opened 2022-09-16
fix(DividerBlock): adjust labels for divider style

Walkeezy:
Ticket: https://app.clickup.com/t/2t7u1ww

### DIFF
--- a/packages/divider-block/src/settings.ts
+++ b/packages/divider-block/src/settings.ts
@@ -200,15 +200,15 @@ export const settings: BlockSettings = {
                     choices: [
                         {
                             value: DividerStyle.Solid,
-                            label: DividerStyle.Solid,
+                            label: 'Solid',
                         },
                         {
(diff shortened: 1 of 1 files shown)


rawi96 (approved):

packages/divider-block/src/settings.ts
```diff
@@ -200,15 +200,15 @@ export const settings: BlockSettings = {
                     choices: [
                         {
                             value: DividerStyle.Solid,
-                            label: 'Solid',
+                            label: DividerStyle.Solid,
```
SamuelAlev, line 203: DividerStyle represent the values not the labels, it is pure luck that they match (if we add word with space then we will have a weird PascalCase label), do you really want to have this change?

Walkeezy: Fair point, changed the label to a string :)

SamuelAlev (approved):

PR #350, opened 2022-09-20
feat(quote-block): add alignment settings

rawi96:
Adds option to align text in the quote block + distinguish between hug text and full-width text
https://app.clickup.com/t/2u5zqkq


Walkeezy (approved):
Very nice

celine-s (commented):
Nice! 🥳

celine-s (approved):

PR #354, opened 2022-09-22
fix(ColorKitBlock): memoize palette ids

leko91:
We have an infinite loop when fetching all color palettes. Ideally we fix this in the app-bridge at some point, but for now memoization should do the job.


GoranRibic (approved):

leko91:
Merged in the https://github.com/Frontify/guideline-blocks/pull/349!

PR #355, opened 2022-09-26
fix(CodeSnippetBlock): design review fixes

leko91:
Tagging @GoranRibic for visibility


packages/shared/src/settings/margin.ts
```diff
@@ -0,0 +1,73 @@
...
+import { MARGIN_DEFAULT_PLACEHOLDER } from './defaultValues';
+import { marginStyleMap, Margin } from './types';
+
+/**
+ * Returns margin settings: margin switch, margin slider, custom margin input
+ *
+ * @param options Options for the settings
+ * @param options.id Custom suffix for the setting ids
+ * @returns {SettingBlock} Returns margin settings
+ */
```
SamuelAlev, line 17: You need to move it on top of `export const getMarginSlider` otherwise it has no effect

packages/shared/src/settings/marginExtended.ts
```diff
@@ -0,0 +1,90 @@
...
+ * @param options Options for the settings
+ * @param options.id Custom suffix for the setting ids
+ * @returns {SettingBlock} Returns margin settings
+ */
+
+type MarginSettingsType = {
+    id?: string;
+};
+
+export const getMarginExtendedSettings = (options?: MarginSettingsType): SettingBlock => {
```
SamuelAlev, line 25: Same for the other files

SamuelAlev (approved):

PR #356, opened 2022-09-26
new color scale block with simplify data handling

dalder:


SamuelAlev:
Little bit outdated ;)

PR #361, opened 2022-09-27
feat(ColorKitBlock): design review changes

leko91:
Based on the comments from here: https://app.clickup.com/t/2wmf7nk

### DIFF
--- a/packages/color-kit-block/src/Color.tsx
+++ b/packages/color-kit-block/src/Color.tsx
@@ -44,7 +44,6 @@ export const Color = ({ isEditing, color }: ColorProps): ReactElement => {
         <Tooltip
             withArrow
             key={color.id}
-            hoverDelay={0}
             position={TooltipPosition.Right}
             content={<TooltipContent colorValue={color.hex} status={status} />}
             triggerElement={
--- a/packages/color-kit-block/src/Color.tsx
+++ b/packages/color-kit-block/src/Color.tsx
@@ -25,38 +25,39 @@ export const Color = ({ isEditing, color }: ColorProps): ReactElement => {
     }
 
     const ColorBox = () => (
-        <div
-            className={merge([
-                "tw-bg-[url('https://cdn.frontify.com/img/transparent.png')] tw-bg-[length:10px_10px]",
-                isEditing &&
-                    '[&:first-child>div]:tw-shadow-inner-line-first [&:last-child>div]:tw-shadow-inner-line-last',
-            ])}
-        >
+        <div className="tw-bg-[url('https://cdn.frontify.com/img/transparent.png')] tw-bg-[length:10px_10px]">
             <div
                 key={color.id}
                 data-test-id="color"
                 style={{
                     backgroundColor: toRgbaString(colorWithDecimalAlpha),
                 }}
-                className="tw-w-6 tw-h-6 tw-overflow-hidden tw-shadow-inner-line-y tw-transition-shadow hover:!tw-shadow-inner-line-strong"
+                className={merge([
+                    'tw-w-6 tw-h-6 tw-overflow-hidden tw-shadow-inner-line-y tw-transition-shadow',
+                    !isEditing && 'hover:!tw-shadow-inner-line-strong',
+                ])}
                 onClick={() => copy(`#${color.hex}`)}
             />
         </div>
     );
 
     return isEditing ? (
-        <ColorBox />
+        <div className="[&:first-child>div>div]:tw-shadow-inner-line-first [&:last-child>div>div]:tw-shadow-inner-line-last">
+            <ColorBox />
+        </div>
     ) : (
-        <Tooltip
-            withArrow
-            key={color.id}
-            position={TooltipPosition.Right}
-            content={<TooltipContent colorValue={color.hex} status={status} />}
-            triggerElement={
-                <div>
-                    <ColorBox />
-                </div>
-            }
-        />
+        <div className="[&:first-child>div>div>div>div]:tw-shadow-inner-line-first [&:last-child>div>div>div>div]:tw-shadow-inner-line-last">
+            <Tooltip
+                withArrow
+                key={color.id}
+                position={TooltipPosition.Right}
+                content={<TooltipContent colorValue={color.hex} status={status} />}
+                triggerElement={
+                    <div>
+                        <ColorBox />
+                    </div>
+                }
+            />
+        </div>
     );
 };
--- a/packages/color-kit-block/src/Color.tsx
+++ b/packages/color-kit-block/src/Color.tsx
@@ -2,7 +2,7 @@
 
 import type { ReactElement } from 'react';
 import type { Color as ColorType } from '@frontify/app-bridge';
-import { Tooltip, TooltipPosition, useCopy } from '@frontify/fondue';
+import { Tooltip, TooltipPosition, merge, useCopy } from '@frontify/fondue';
 import { toRgbaString } from '@frontify/guideline-blocks-shared';
 
 import { TooltipContent } from './TooltipContent';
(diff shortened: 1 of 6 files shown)


packages/color-kit-block/src/Palette.tsx
```diff
@@ -1,17 +1,30 @@
...
-import type { PaletteProps } from './types';
+import { EmptyView } from './EmptyView';
 import { Color } from './Color';
+import type { PaletteProps } from './types';
 
 export const Palette = ({ palette, isEditing }: PaletteProps) => {
     const { name, colors = [] } = palette;
 
+    if (colors.length === 0) {
+        return <EmptyView paletteName={`${name || 'Unitled Palette'} (no colors added)`} />;
```
celine-s, line 13:
```suggestion
        return <EmptyView paletteName={`${name ?? 'Untitled Palette'} (no colors added)`} />;
```

packages/color-kit-block/src/Palette.tsx
```diff
@@ -1,17 +1,30 @@
...
+        <div data-test-id="palette" className="tw-flex tw-flex-col tw-space-y-2 tw-mb-3 last:tw-mb-0">
             <Text color="x-weak">{name}</Text>
-            <div className="tw-flex tw-flex-wrap [&>div]:tw-h-6 [&>div:first-child>div]:tw-shadow-inner-line-first [&>div:last-child>div]:tw-shadow-inner-line-last">
+
+            <div
+                className={merge([
+                    'tw-flex tw-flex-wrap',
+                    isEditing
+                        ? '[&>div:first-child>div]:tw-shadow-inner-line-first [&>div:last-child>div]:tw-shadow-inner-line-last'
+                        : '[&>div:first-child>div>div>div]:tw-shadow-inner-line-first [&>div:last-child>div>div>div]:tw-shadow-inner-line-last',
```
celine-s, line 25: `[&>div:first-child>div>div>div]` :D

packages/color-kit-block/src/EmptyView.tsx
```diff
@@ -0,0 +1,27 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { Text } from '@frontify/fondue';
+
+type EmptyViewProps = {
+    paletteName?: string;
+};
+
+export const EmptyView = ({ paletteName = 'Select color palettes for them to appear here' }: EmptyViewProps) => {
```
celine-s, line 9: wdyt? or is this the design?
```suggestion
export const EmptyView = ({ paletteName = 'Select color palettes in the settings for them to appear here' }: EmptyViewProps) => {
```

celine-s (changes requested):
except that the name of an untitled Palette is not shown in the UI (see comment with screenshot), everything works fine 👍

packages/color-kit-block/src/EmptyView.tsx
leko91: This is the current copy, not my call to change it 😄

packages/color-kit-block/src/Palette.tsx
leko91: I can explain in a call, this is the only way 😭

packages/color-kit-block/src/Palette.tsx
leko91: The idea was to have `||` because when the name is an empty string, we have `Untitled Palette` as a title. With `??` we would just get an empty string and a wrong palette name.

packages/color-kit-block/src/EmptyView.tsx
celine-s: 👍 :)

packages/color-kit-block/src/Palette.tsx
celine-s: I was just wondering if you could take it within the `<Color/>` component, such that it's clearer where you want to have your style :D. Probably you need to change style of the fondue component `Tooltip`.?

packages/color-kit-block/src/Palette.tsx
celine-s: true that . keep || 👍

packages/color-kit-block/src/Palette.tsx
leko91: Exactly, I cannot access the element I want, and we cannot add classes to `Tooltip` so this is what we get 😄

celine-s: fair enough 😄
maybe you could add a <div> around the <Tooltip/> component with this styling such that it's clearer where it belongs?

leko91: I don't see the message, add a what? 😄

packages/color-kit-block/src/Palette.spec.ct.tsx
```diff
@@ -33,7 +34,7 @@ describe('Palette', () => {
     it('should render palette without colors', () => {
         mount(<Palette palette={{ ...dummyPalette, colors: [] }} isEditing={false} />);
 
-        cy.get(ColorKitPaletteSelector).should('not.have.descendants', ColorKitColorSelector);
+        cy.get(EmptyViewSelector).should('exist');
```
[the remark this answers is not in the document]
leko91: There is a test that checks the rendered colors. Or do you mean the placeholder colors?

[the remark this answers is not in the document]
celine-s: I meant the placeholder. sorry could have been more specific. :D

packages/color-kit-block/src/Palette.tsx
celine-s: wtf 😂 Now it's fixed and you should see it

packages/color-kit-block/src/Color.tsx
```diff
@@ -32,26 +32,32 @@ export const Color = ({ isEditing, color }: ColorProps): ReactElement => {
...
+                    !isEditing && 'hover:!tw-shadow-inner-line-strong',
+                ])}
                 onClick={() => copy(`#${color.hex}`)}
             />
         </div>
     );
 
     return isEditing ? (
-        <ColorBox />
+        <div className="[&:first-child>div>div]:tw-shadow-inner-line-first [&:last-child>div>div]:tw-shadow-inner-line-last">
```
celine-s, line 45: extract to const className="[&:first-child>div>div]:tw-shadow-inner-line-first [&:last-child>div>div]:tw-shadow-inner-line-last"
 and pass that className
```suggestion
        <div className={className}>
```

packages/color-kit-block/src/Palette.tsx
```diff
@@ -2,16 +2,22 @@
...
-import type { PaletteProps } from './types';
+import { EmptyView } from './EmptyView';
 import { Color } from './Color';
+import type { PaletteProps } from './types';
 
 export const Palette = ({ palette, isEditing }: PaletteProps) => {
     const { name, colors = [] } = palette;
 
+    if (colors.length === 0) {
+        return <EmptyView paletteName={`${name || 'Untitled Palette'} (no colors added)`} />;
```
celine-s, line 13: could you just pass name? And then add the rest within the EmptyView?

celine-s (approved):

packages/color-kit-block/src/ColorKitBlock.tsx
```diff
@@ -44,14 +46,21 @@ export const ColorKitBlock = ({ appBridge }: ColorKitBlockProps): ReactElement =
                     title="download color palettes"
                     style={{ pointerEvents: isDownloadEnabled ? 'initial' : 'none' }}
                 >
-                    <Button style={ButtonStyle.Secondary} icon={<IconArrowCircleDown />} disabled={!isDownloadEnabled}>
+                    <Button
+                        emphasis={ButtonEmphasis.Default}
```
SamuelAlev, line 50: Is it needed if it is the default? 🤔

leko91: Actually yes 😄

SamuelAlev: 😶‍🌫️

PR #364, opened 2022-09-29
feat(CodeSnippetBlock): set clearable settings to 0px

leko91:


packages/shared/src/settings/borderRadiusExtended.ts
```diff
@@ -58,28 +58,52 @@ export const getExtendedBorderRadiusSettings = (options?: BorderRadiusSettingsTy
...
+                            } else {
+                                appendUnit(bundle, topLeftId);
+                            }
+                        },
                     },
                     {
                         id: topRightId,
                         type: 'input',
                         label: 'Top Right',
                         rules: [numericalOrPixelRule],
```
SamuelAlev, line 73: Actually not "approved" because it collides with the `numericalOrPixelRule` rule

SamuelAlev (changes requested):

packages/shared/src/settings/borderRadiusExtended.ts
leko91: I think that this should still be correct.
I clear the input -> it gets set to 0px -> I clear again -> field is empty but incorrect. Do we want to allow the user to save an empty string?

PR #366, opened 2022-09-29
feat: add settings and type for block with image

celine-s:


packages/color-matching-block/src/settings.tsx
```diff
@@ -0,0 +1,124 @@
...
+export const ICON_RIGHT_ID = 'iconRight';
+export const LOGO_ID = 'logoId';
+export const IS_TOOLBAR_TYPE = 'toolbarType';
+
+export const settings = {
+    main: [
+        {
+            id: IS_TOOLBAR_TYPE,
+            type: 'dropdown',
+            size: 'Large' as DropdownSize.Large,
```
fulopdaniel, line 20: Instead of casting is it possible to simply do
`size: DropdownSize.Large` ?
Same for the svg ones below

packages/color-matching-block/src/types.ts
```diff
@@ -2,22 +2,34 @@
...
+    imageTitle?: string;
+    imageDescription?: string;
+};
+
+export enum ColorMatchingType {
+    Render = 'Render',
+    Toolbar = 'Toolbar',
+}
+
+export type IconProps = {
```
fulopdaniel, line 33: Is this used somewhere? Maybe you need to add to the Icons in foundation?

celine-s: Oh good catch :D I've used them in other icons. I added them now also to those icons & moved it to the foundation folder

fulopdaniel (approved):

PR #374, opened 2022-10-13
fix(ColorKitBlock): Force Text component to be display block

leko91:


GoranRibic (approved):

packages/color-kit-block/src/Palette.tsx
```diff
@@ -14,7 +14,7 @@ export const Palette = ({ palette, isEditing }: PaletteProps) => {
     }
 
     return (
-        <div data-test-id="palette" className="tw-space-y-2 tw-mb-3 last:tw-mb-0">
+        <div data-test-id="palette" className="tw-space-y-2 tw-mb-3 last:tw-mb-0 [&>span]:tw-block">
             <Text color="x-weak">{name || 'Untitled Palette'}</Text>
```
SamuelAlev, line 18:
```suggestion
        <div data-test-id="palette" className="tw-space-y-2 tw-mb-3 last:tw-mb-0">
            <Text color="x-weak" as="p">{name || 'Untitled Palette'}</Text>
```
Could it also work?

SamuelAlev: Oh if it is only on beta then that is just some old blocks messing around with global styling and you shouldn't have to fix anything 
Let me ask the G1 team to update the blocks

leko91: Did you check with them? Not sure what the course of action is here

PR #380, opened 2022-10-18
fix: improve callout block settings

mike-schmid:
Ticket: https://app.clickup.com/t/2ry3j8m

### DIFF
--- a/packages/callout-block/src/settings.ts
+++ b/packages/callout-block/src/settings.ts
@@ -28,14 +28,9 @@ export const settings: BlockSettings = {
             size: 'Large' as DropdownSize.Large,
             choices: [
                 {
-                    value: Type.Warning,
-                    icon: 'ExclamationMarkTriangle' as IconEnum.ExclamationMarkTriangle,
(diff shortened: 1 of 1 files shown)


packages/callout-block/src/settings.ts
```diff
@@ -24,7 +24,7 @@ export const settings: BlockSettings = {
         {
             id: 'type',
             type: 'dropdown',
-            defaultValue: Type.Warning,
+            defaultValue: Type.Info,
```
Walkeezy, line 27: I would also move `Info` to the top of the dropdown (also like this in the design).

triggertoo (approved):
looks good 👍

packages/callout-block/src/settings.ts
mike-schmid: Good point. I've updated the order. Can you please approve again?
thanks

mike-schmid:
close this PR and make a new one because of unverified commits.
new PR: https://github.com/Frontify/guideline-blocks/pull/382

PR #391, opened 2022-10-31
refactor: move colors-block to color-block

SamuelAlev:
~~TODO: Remove the need of `window.blockSettings[appBridge.getBlockId()]['colorPaletteId'] = colorPalette.id`~~

**Why is it needed?**
`useBlockSettings` set the initial value out of the `window` object, maybe we need to update App Bridge `updateBlockSettings` to do it?


triggertoo (approved):
👍 tested with https://github.com/Frontify/clarify/pull/11525

PR #395, opened 2022-11-01
feat(CalloutBlock): use colors from guidelines design tokens

Walkeezy:
- Extend useGuidelineDesignTokens to get Callout colors
- Extend isDark shared utility to accept other inputs (e.g. hex & rgb strings) than type `Color`
- Use colors from Guidelines Design Tokens in Callout block
- Switch text color black/white based on background color in Callout block


packages/shared/src/utilities/color/isDark.ts
```diff
@@ -10,7 +10,14 @@ import { toShortRgba } from './toShortRgba';
...
+
+const isRgbaLongFormat = (value: Color) => {
+    const requiredKeys = ['red', 'green', 'blue'];
+    return typeof value === 'object' && requiredKeys.every((i) => value.hasOwnProperty(i));
+};
+
+export const isDark = (color: unknown): boolean => {
+    const inputColor = isRgbaLongFormat(color as Color) ? toShortRgba(color) : (color as ColorInput);
+    const parsedColor = tinycolor(inputColor);
     return parsedColor.isDark() || (parsedColor.getAlpha() > 0.25 && parsedColor.getAlpha() < 1);
```
mike-schmid, line 22: Just FYI: there is a function in the tinycolor library to get the most readable color: https://www.npmjs.com/package/tinycolor2#mostreadable

mike-schmid, line 22: It seems that tinycolor cannot handle our color format correctly. isDark always returns "true" even if you select a white background for warnings.

image

No matter what you select as background color, the font color is always white...

mike-schmid (changes requested):

celine-s (approved):
Nice! 👍 

The color should be black if the bg is white, right? (but I think Mike already pointed it out)

packages/shared/src/utilities/color/isDark.ts
Walkeezy: Nice! But I'd rather keep it as is, because `isDark` now returns a boolean and `mostReadable` returns you the most readable color value. Don't see a benefit in using that.

Walkeezy: Good catch! Fixed.

mike-schmid (approved):

PR #401, opened 2022-11-07
feat(CalloutBlock): add appearance

celine-s:


packages/callout-block/src/CalloutBlock.tsx
```diff
@@ -33,19 +48,24 @@ export const CalloutBlock: FC<CalloutBlockProps> = ({ appBridge }) => {
...
-    ]);
-
     const customPaddingStyle = {
         padding: blockSettings.hasCustomPadding
             ? `${blockSettings.paddingTop} ${blockSettings.paddingRight} ${blockSettings.paddingBottom} ${blockSettings.paddingLeft}`
             : '',
     };
 
+    const color = getBackgroundColor(blockSettings.type);
+    const backgroundColor = blockSettings.appearance === Appearance.Strong ? setAlpha(color, 1) : setAlpha(color, 0.1);
```
Walkeezy, line 58: This way the color of the user, if has selected a alpha value, is overwritten. Is this correct? If not, you could do it this way and it would keep the users color:

```suggestion
    const backgroundColor = blockSettings.appearance === Appearance.Strong ? color : setAlpha(color, 0.1);
```

packages/callout-block/src/CalloutBlock.tsx
```diff
@@ -33,19 +48,24 @@ export const CalloutBlock: FC<CalloutBlockProps> = ({ appBridge }) => {
...
+
+    const defaultTextColor = isDark(color) ? 'white' : 'black';
+    const textColor = blockSettings.appearance === Appearance.Light ? setAlpha(color, 1) : defaultTextColor;
```
Walkeezy, line 61: Same here, like this the alpha value of the user (global design tokens) is ignored.

packages/shared/src/utilities/color/isRgbaLongFormat.ts
```diff
@@ -0,0 +1,6 @@
+import { Color } from '@frontify/fondue';
```
Walkeezy, line 1:
```suggestion
/* (c) Copyright Frontify Ltd., all rights reserved. */

import { Color } from '@frontify/fondue';
```

packages/shared/src/utilities/color/isRgbaLongFormat.ts
```diff
@@ -0,0 +1,6 @@
...
+
+export const isRgbaLongFormat = (value: Color) => {
```
Walkeezy, line 3: Nice 👍🏻  could you also use this in the `toShortRgba` utility? And maybe add some tests?

packages/callout-block/src/CalloutBlock.tsx
mike-schmid: Another thing. I think currently it's a little confusing that `getBackgroundColor(...)` returns something of type `string | undefined`, the parameter of setAlpha is typed as `color: unknown` and in its implementation we just assume its of type Color or ColorInput `isRgbaLongFormat(color as Color) ? toShortRgba(color) : (color as ColorInput)`

mike-schmid, line 58: Another thing. I think currently it's a little confusing that `getBackgroundColor(...)` returns something of type `string | undefined`, the parameter of `setAlpha(...)` is typed as `color: unknown` and in its implementation we just assume its of type Color or ColorInput `isRgbaLongFormat(color as Color) ? toShortRgba(color) : (color as ColorInput)`
Can't we fix that typing?

mike-schmid (changes requested):

packages/callout-block/src/CalloutBlock.tsx
celine-s: you're right. I changed it to your suggestion - thanks 👍

packages/shared/src/utilities/color/isRgbaLongFormat.ts
celine-s: 👍

celine-s: removed the file again, as I actually don't need the isRgbaLongFormat.ts in this case

packages/shared/src/utilities/color/isRgbaLongFormat.ts
celine-s: removed the file again, as I actually don't need the isRgbaLongFormat.ts in this case

mike-schmid (approved):

Walkeezy (approved):

PR #402, opened 2022-11-08
feat: new color scale block

dominiquemb:
Description of changes:

- Previously, after adding 2 colors, the 3rd color and those after it would have a width of 18px (MINIMUM_COLOR_WIDTH*2).  Simone explained that every color should take up an equal amount of space in the block, except if it has been resized manually.  

For example, if 2 colors have been added, they should each take up 50% of the block width.  If a user adds a 3rd color, then every color should take up 1/3 of the width of the block.

If the user resizes color #2 to take up 80% of the block, then this color block should be left as-is if any new colors are added. When color #3 added, it should take up 10% of the block width, and color #1, should get automatically resized to 10% width, as well.

- In the Settings panel, when focus is removed from the custom height input field, the string 'px' should be added to any value in the field.


packages/color-scale-block/src/settings.ts
```diff
@@ -39,6 +39,13 @@ export const settings: BlockSettings = {
                     defaultValue: '100px',
                     clearable: false,
                     rules: [numericalOrPixelRule],
+                    onChange: (bundle: Bundle): void => {
+                        console.log(bundle.getBlock('heightInput')?.value);
```
SamuelAlev, line 43:
```suggestion
```

packages/color-scale-block/src/ColorScaleBlock.tsx
```diff
@@ -323,7 +378,7 @@ export const ColorScaleBlock = ({ appBridge }: ColorScaleBlockProps) => {
                     style={{
                         height: colorScaleHeight,
                     }}
-                    className={'tw-overflow-hidden tw-rounded tw-flex tw-max-w-full'}
+                    className={'tw-rounded tw-flex tw-max-w-full'}
```
SamuelAlev, line 381:
```suggestion
                    className="tw-rounded tw-flex tw-max-w-full"
```

packages/color-scale-block/src/ColorScaleBlock.tsx
```diff
@@ -180,6 +191,46 @@ export const ColorScaleBlock = ({ appBridge }: ColorScaleBlockProps) => {
...
     };
 
+    const calculateNewWidthForUnresizedColors = (colorArray: ColorProps[]) => {
+        if (colorScaleBlockRef.current === null) {
+            return;
+        }
+
+        const unresizedColors = colorArray.filter((color) => !color.resized);
+
+        const numberOfUnresizedColors = unresizedColors.length ?? 0;
```
SamuelAlev, line 201:
```suggestion
        const numberOfUnresizedColors = unresizedColors.length;
```

packages/color-scale-block/src/ColorScaleBlock.tsx
```diff
@@ -180,6 +191,46 @@ export const ColorScaleBlock = ({ appBridge }: ColorScaleBlockProps) => {
...
+            }
+            return accumulator;
+        }, 0);
+
+        const resizedColorsTotalWidth = colorArray.reduce((accumulator, color) => {
+            if (color.resized) {
+                return accumulator + color.width + COLOR_SQUARE_SPACING;
+            }
+            return accumulator;
+        }, 0);
```
SamuelAlev, line 215: Can you merge them into one loop?

packages/color-scale-block/src/ColorScaleBlock.tsx
```diff
@@ -180,6 +191,46 @@ export const ColorScaleBlock = ({ appBridge }: ColorScaleBlockProps) => {
...
+        const spaceToDivide =
+            unresizedColorsTotalWidth > blockWidth
+                ? blockWidth - resizedColorsTotalWidth
+                : unresizedColorsTotalWidth + emptySpace;
+
+        const calculatedNewWidth = colorArray.length > 0 ? spaceToDivide / numberOfUnresizedColors : blockWidth;
+
+        return calculatedNewWidth >= MINIMUM_COLOR_WIDTH
+            ? calculatedNewWidth - COLOR_SQUARE_SPACING
+            : MINIMUM_COLOR_WIDTH - COLOR_SQUARE_SPACING;
```
SamuelAlev, line 231: Can you make it more readable?

dominiquemb:
Description of changes:

- Previously, after adding 2 colors, the 3rd color and those after it would have a width of 18px (MINIMUM_COLOR_WIDTH * 2).  Simone explained that every color should take up an equal amount of space in the block, except if it has been resized manually.  

For example, if 2 colors have been added, they should each take up 50% of the block width.  If a user adds a 3rd color, then every color should take up 1/3 of the width of the block.

If the user resizes color #2 to take up 80% of the block, then this color block should be left as-is if any new colors are added. When color #3 added, it should take up 10% of the block width, and color #1, should get automatically resized to 10% width, as well.

- In the Settings panel, when focus is removed from the custom height input field, the string 'px' should be added to any value in the field.

packages/color-scale-block/src/settings.ts
```diff
@@ -39,6 +39,13 @@ export const settings: BlockSettings = {
                     defaultValue: '100px',
                     clearable: false,
                     rules: [numericalOrPixelRule],
+                    onChange: (bundle: Bundle): void => {
```
SamuelAlev, line 42:
```suggestion
                    onChange: (bundle) => {
```
Type are defined by `BlockSettings`

dominiquemb:
@SamuelAlev It seems like there are 3 tests failing, and one of them has to do with the resizing feature. Nothing in the test has changed, but it's no longer working. When I test the resize feature manually, it works fine. But with Cypress, it seems like the cursor movement is not happening.  I've tried a variety of different approaches: `realMouseDown`, `realMouseMove`, `drag(target)`, `move({deltaX: some_number})`, `trigger('mousedown')`, `trigger('mousemove'` but none of them work anymore.  

Is it okay if I just remove this test or comment it out for now?

GoranRibic (approved):

PR #403, opened 2022-11-08
fix(PersonalNoteBlock): check for valid date

Walkeezy:
Check if date actually is valid, output string otherwise.

### DIFF
--- a/packages/personal-note-block/src/components/NoteHeader.tsx
+++ b/packages/personal-note-block/src/components/NoteHeader.tsx
@@ -26,7 +26,7 @@ export const NoteHeader: FC<NoteHeaderProps> = ({
             {hasAvatarName && <span className={useLightText ? 'tw-text-white' : 'tw-text-black'}>{name}</span>}
             {hasDateEdited && dateEdited && (
                 <span className={useLightText ? 'tw-text-white tw-opacity-60' : 'tw-text-x-weak'}>
-                    Last edited on {dayjs(dateEdited).format('L, LT')}
+                    Last edited on {dayjs(dateEdited).isValid() ? dayjs(dateEdited).format('L LT') : dateEdited}
                 </span>
             )}
         </div>


packages/personal-note-block/src/components/NoteHeader.tsx
```diff
@@ -26,7 +26,7 @@ export const NoteHeader: FC<NoteHeaderProps> = ({
             {hasAvatarName && <span className={useLightText ? 'tw-text-white' : 'tw-text-black'}>{name}</span>}
             {hasDateEdited && dateEdited && (
                 <span className={useLightText ? 'tw-text-white tw-opacity-60' : 'tw-text-x-weak'}>
-                    Last edited on {dayjs(dateEdited).format('L, LT')}
+                    Last edited on {dayjs(dateEdited).isValid() ? dayjs(dateEdited).format('L LT') : dateEdited}
```
SamuelAlev, line 29: Why `dayjs` and not `Intl.DateTimeFormat`? https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat

Walkeezy: We get many different date formats from the old blocks with the migration, so dayjs is able to parse (hopefully) most of them.

celine-s (approved):
I think we should still handle the dates which are invalid. Or shouldn't we? This way most of the dates which are in the old blocks are invalid. e.g. those are invalid dates:
- 24.7.2019, 15:57:37
- 3\/2\/2016, 10:04:54 PM
- 30. September 2015 19:46:51 MESZ
- 8\/19\/2019, 2:10:58 PM
* 23. Mai 2016 um 16:21:47 MESZ
* 8. August 2016 11:43:51 MESZ
* 17. März 2016 um 19:00:36 MEZ

packages/personal-note-block/src/components/NoteHeader.tsx
SamuelAlev: Then migrate the date during migration so we have a great baseline to start

PR #404, opened 2022-11-08
feat(CalloutBlock): add predefined icons

celine-s:
open PRs to enable icon & label within slider:
- [clarify](https://github.com/Frontify/clarify/pull/11963): enable mixed slider values (icon & label)
- [fondue](https://github.com/Frontify/fondue/pull/1032): adapt props

### DIFF
--- a/packages/callout-block/src/CalloutBlock.tsx
+++ b/packages/callout-block/src/CalloutBlock.tsx
@@ -12,11 +12,14 @@ import {
 } from '@frontify/guideline-blocks-shared';
 import { FC } from 'react';
 import 'tailwindcss/tailwind.css';
+import { CalloutIcon } from './components/CalloutIcon';
+import { hasRichTextValue } from './utils/hasRichTextValue';
 import { ICON_ASSET_ID } from './settings';
 import {
     Appearance,
     BlockSettings,
     CalloutBlockProps,
+    Icon,
     Type,
     Width,
     alignmentMap,
@@ -88,11 +91,11 @@ export const CalloutBlock: FC<CalloutBlockProps> = ({ appBridge }) => {
                     ...customCornerRadiusStyle,
                 }}
             >
-                {blockSettings.iconSwitch && iconUrl && (
-                    <span className="tw-mr-3 tw-flex-none tw-w-6 tw-h-6">
-                        <img data-test-id="callout-icon" alt="Callout icon" src={iconUrl} />
-                    </span>
-                )}
+                <CalloutIcon
+                    iconType={blockSettings.iconSwitch ? Icon.Custom : blockSettings.iconType}
+                    iconUrl={iconUrl}
+                    hasValue={hasRichTextValue(blockSettings.textValue)}
+                />
                 <RichTextEditor
                     onTextChange={onTextChange}
                     readonly={!isEditing}
--- a/packages/callout-block/src/components/CalloutIcon.tsx
+++ b/packages/callout-block/src/components/CalloutIcon.tsx
@@ -1,0 +1,34 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { IconInfo20, IconLightbulb20, IconMegaphone20 } from '@frontify/fondue';
+import { joinClassNames } from '@frontify/guideline-blocks-shared';
+import { ReactElement } from 'react';
+import { Icon } from '../types';
+
+type CalloutIconProps = {
+    iconType: Icon;
+    hasValue: boolean;
+    iconUrl?: string;
+};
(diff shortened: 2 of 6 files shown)


packages/callout-block/src/components/CalloutIcon.tsx
```diff
@@ -0,0 +1,34 @@
...
+        )}
+    </>
+);
+
+export const calloutIconMap = (iconUrl?: string): Record<Icon, ReactElement> => ({
+    none: <></>,
+    info: <IconInfo20 />,
+    lightbulb: <IconLightbulb20 />,
+    megaphone: <IconMegaphone20 />,
+    custom: <img data-test-id="callout-icon" alt="Callout icon" src={iconUrl} />,
```
mike-schmid, line 33: Decorative icons should have an empty alt text (https://www.w3.org/WAI/tutorials/images/decorative/).
It's no use for a blind person if a screen reader says "Callout Icon" :)

packages/callout-block/src/utils/hasRichTextValue.ts
```diff
@@ -0,0 +1,21 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
Walkeezy, line 3: Use `// eslint-disable-next-line @typescript-eslint/no-explicit-any` for the lines, instead of disable it for the whole file.

packages/callout-block/src/components/CalloutIcon.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { joinClassNames } from '@frontify/guideline-blocks-shared';
+import { ReactElement } from 'react';
+import { Icon } from '../types';
+
+type CalloutIconProps = {
+    iconType: Icon;
+    hasValue: boolean;
+    iconUrl?: string;
+};
+export const CalloutIcon = ({ iconType, hasValue, iconUrl }: CalloutIconProps) => (
```
Walkeezy, line 13: Nit: To me the naming of `hasValue` seems a bit confusing. `isActive` or something similiar would probably make more sense.

packages/callout-block/src/utils/hasRichTextValue.ts
```diff
@@ -0,0 +1,21 @@
...
+export const hasRichTextValue = (string?: string) => {
+    if (!string) {
+        return false;
+    }
+    const json = JSON.parse(string);
```
mike-schmid, line 8: this line will crash if the passed string is not a valid json (for example "`asdf`"). Can this happen or is it always a valid json?

packages/callout-block/src/CalloutBlock.tsx
```diff
@@ -88,11 +91,11 @@ export const CalloutBlock: FC<CalloutBlockProps> = ({ appBridge }) => {
...
-                {blockSettings.iconSwitch && iconUrl && (
-                    <span className="tw-mr-3 tw-flex-none tw-w-6 tw-h-6">
-                        <img data-test-id="callout-icon" alt="Callout icon" src={iconUrl} />
-                    </span>
-                )}
+                <CalloutIcon
+                    iconType={blockSettings.iconSwitch ? Icon.Custom : blockSettings.iconType}
+                    iconUrl={iconUrl}
+                    hasValue={hasRichTextValue(blockSettings.textValue)}
+                />
```
Walkeezy, line 98: I would keep the condition to show the icon outside of the `CalloutIcon` component. If there is no icon, we also don't need to run `hasRichTextValue`.

mike-schmid (approved):

packages/callout-block/src/CalloutBlock.tsx
```diff
@@ -73,6 +76,7 @@ export const CalloutBlock: FC<CalloutBlockProps> = ({ appBridge }) => {
     };
 
     const iconUrl = blockSettings.iconSwitch ? blockAssets?.[ICON_ASSET_ID]?.[0]?.genericUrl : '';
+    const hasNoIcon = blockSettings.iconType === Icon.None || (blockSettings.iconType === Icon.Custom && !iconUrl);
```
Walkeezy, line 79: Untested, but `blockSettings.iconType` seems to be undefined when a new block is created so you would also need to check for that:
```suggestion
    const hasNoIcon = !blockSettings.iconType || blockSettings.iconType === Icon.None || (blockSettings.iconType === Icon.Custom && !iconUrl);
```

Currently there is an empty space where the icon should be if you create a new block. If you then switch to Icon "none" the space disappears.

Or just set the default iconType to None.

celine-s: True! Thanks for checking again! Fixed

PR #405, opened 2022-11-09
feat(ChecklistBlock): fixes & improvements

Walkeezy:
Checklist block fixes from [this ticket](https://app.clickup.com/t/2ry3j9p): 

- Use body font style for checklist item
- Use 'image caption' font style for date in checklist item
- Adjust settings according to Figma design
- Adjust default color values according to Figma design
- Use small switches in settings
- Switch default type to "Checkbox"
- Remove padding settings
- Adjust style for controls buttons (move & delete item)
- Prevent HTML & formatted text to be pasted as input for a checklist item


mike-schmid (approved):

packages/shared/src/hooks/useGuidelineDesignTokens.ts
```diff
@@ -44,7 +44,8 @@ export type DesignTokenName =
     | 'button_primary'
     | 'button_secondary'
     | 'button_tertiary'
-    | 'callout';
+    | 'callout'
+    | 'image-caption';
```
triggertoo, line 48: i'd suggest wo use camel case for all tokens and map it accordingly, wdyt?

triggertoo (approved):

PR #407, opened 2022-11-09
fix(CodeSnippetBlock): replace border setting by using predefined getBorderSettings()

celine-s:


Walkeezy (approved):

packages/code-snippet-block/src/CodeSnippetBlock.spec.ct.tsx
```diff
@@ -80,9 +80,9 @@ it('renders code snippet with line numbers', () => {
 it('renders code snippet with border', () => {
     const [CodeSnippetWithStubs] = withAppBridgeBlockStubs(CodeSnippetBlock, {
         blockSettings: {
-            withBorder: true,
-            lineWidth: '2px',
-            lineStyle: 'solid',
+            hasBorder: true,
+            borderWidth: '2px',
+            borderStyle: 'solid',
```
mike-schmid, line 85: these 2 props should also be changes in the next 2 tests  below :)

Walkeezy: Or they could even be removed in the other tests, because the border setting isn't event tested in those tests 😅

packages/code-snippet-block/src/CodeSnippetBlock.spec.ct.tsx
```diff
@@ -82,20 +82,20 @@ it('renders code snippet with border', () => {
...
+            borderStyle: 'Solid',
             borderColor: EXAMPLE_COLOR,
         },
     });
     mount(<CodeSnippetWithStubs />);
     cy.get(codeSnippetEditorSelector).should('have.css', 'border', '2px solid rgb(22, 181, 181)');
 });
 
-it('renders code snippet with border radius', () => {
+it.only('renders code snippet with border radius', () => {
```
mike-schmid, line 93: only?

celine-s: 🤫 removed

packages/code-snippet-block/src/CodeSnippetBlock.spec.ct.tsx
```diff
@@ -109,9 +109,9 @@ it('renders code snippet with border radius', () => {
 it('renders code snippet with custom border radius', () => {
     const [CodeSnippetWithStubs] = withAppBridgeBlockStubs(CodeSnippetBlock, {
         blockSettings: {
-            withBorder: true,
-            lineWidth: '2px',
-            lineStyle: 'solid',
+            hasBorder: true,
+            borderWidth: '2px',
+            borderStyle: 'Solid',
```
Walkeezy, line 114: Can be removed, not tested in the test :)

celine-s: I thought I leave it as then it's clearer where to the radius is set when looking at it again. But actually you're right. Removed it 👍

mike-schmid (approved):

PR #418, opened 2022-11-16
feat(CodeSnippetBlock): revamp

Walkeezy:
Rebuild block with `@uiw/react-codemirror` as packages used before are deprecated.

No support for those languages from the old block:
- HAML
- JSP
- Pug (Jade)
- Slim

### DIFF
--- a/packages/code-snippet-block/src/CodeSnippetBlock.tsx
+++ b/packages/code-snippet-block/src/CodeSnippetBlock.tsx
@@ -78,9 +78,7 @@ export const CodeSnippetBlock = ({ appBridge }: CodeSnippetProps): ReactElement 
                     data-test-id="code-snippet-header"
                     className="tw-py-2 tw-px-3 tw-bg-black-5 tw-border-b tw-border-black-10 tw-text-s"
                 >
-                    {!isEditing && <span>{languageNameMap[selectedLanguage]}</span>}
-
-                    {isEditing && (
+                    {isEditing ? (
                         <div className="tw-max-w-[150px]">
                             <Dropdown
                                 size={DropdownSize.Small}
@@ -97,6 +95,8 @@ export const CodeSnippetBlock = ({ appBridge }: CodeSnippetProps): ReactElement 
                                 onChange={(value) => handleLanguageChange(value as Language)}
                             />
                         </div>
+                    ) : (
+                        <span>{languageNameMap[selectedLanguage]}</span>
                     )}
                 </div>
             )}


packages/code-snippet-block/src/CodeSnippetBlock.tsx
```diff
@@ -1,53 +1,127 @@
...
     } = blockSettings;
 
+    const getTheme = () => {
+        if (theme === 'default') {
+            return 'light';
+        } else if (Object.keys(themes).includes(theme)) {
+            return themes[theme];
+        }
+        return 'light';
+    };
```
mike-schmid, line 42: Couldn't this be simplified to 
`const getTheme = () => themes[theme] ?? 'light';`

packages/code-snippet-block/src/CodeSnippetBlock.tsx
```diff
@@ -1,53 +1,127 @@
...
+
+    const getCurrentLanguageFromLangs = () => {
+        if (selectedLanguage === 'plain') {
+            return null;
+        }
+        if (Object.keys(langs).includes(selectedLanguage)) {
+            return langs[selectedLanguage];
+        }
+        return langs.html;
```
mike-schmid, line 51: shouldn't the default be plain text as well?

packages/code-snippet-block/src/CodeSnippetBlock.tsx
```diff
@@ -1,53 +1,127 @@
...
+    };
```
triggertoo, line 52:
```suggestion
    const currentLanguage = langs[selectedLanguage];
```

packages/code-snippet-block/src/CodeSnippetBlock.tsx
```diff
@@ -1,53 +1,127 @@
...
+                                        menuItems: Object.entries(languageNameMap).map(([value, label]) => ({
+                                            id: value,
+                                            title: label,
+                                        })),
+                                    },
+                                ]}
+                                onChange={(value) => handleLanguageChange(value as Language)}
+                            />
+                        </div>
+                    )}
```
triggertoo, line 105:
```suggestion
                    {isEditing ? (
                        <div className="tw-max-w-[150px]">
                            <Dropdown
                                size={DropdownSize.Small}
                                activeItemId={selectedLanguage}
                                menuBlocks={[
                                    {
                                        id: 'languages',
                                        menuItems: Object.entries(languageNameMap).map(([value, label]) => ({
                                            id: value,
                                            title: label,
                                        })),
                                    },
                                ]}
                                onChange={(value) => handleLanguageChange(value as Language)}
                            />
                        </div>
                    ) : <span>{languageNameMap[selectedLanguage]}</span>}
```

packages/code-snippet-block/src/CodeSnippetBlock.tsx
Walkeezy: `theme` can not be used as an index for `themes`, because `default` does not exist in the `Themes` type, so I had to do it this way.

Walkeezy: But I changed it to one condition only :)

packages/code-snippet-block/src/CodeSnippetBlock.tsx
Walkeezy: Changed 👍🏻

packages/code-snippet-block/src/CodeSnippetBlock.tsx
Walkeezy: Needs to be conditional because `plain` doesn't exist in `langs`, but refactored it a bit.

mike-schmid (approved):

triggertoo (approved):

PR #419, opened 2022-11-18
feat(QuoteBlock): update icons

Walkeezy:
Update icons to be the same as in the old block (as discussed with Schirmi & Martin) and fix an issue, where the icon appears very big if a custom value is cleared.


packages/quote-block/src/QuoteBlock.tsx
```diff
@@ -68,7 +77,11 @@ export const QuoteBlock: FC<Props> = ({ appBridge }) => {
                         color={iconColor}
                         blockAssets={blockAssets}
                         isCustomSize={blockSettings.isCustomSize}
-                        sizeValue={blockSettings.sizeValue}
+                        sizeValue={
+                            blockSettings.sizeValue && blockSettings.sizeValue !== ''
```
mike-schmid, line 81: isn't `blockSettings.sizeValue !== ''` redundant anyway?

Walkeezy: Ah true, thanks. Fixed!

mike-schmid (approved):

rawi96 (approved):

PR #422, opened 2022-11-21
feat(FigmaBlock): various improvements

fulopdaniel:
Improvements:

- Min height changed to 50
- Responsive on mobile
- Added advanced border options
- Live mode also responds to style and height mods


bojangles-m (approved):

packages/figma-block/src/settings.ts
```diff
@@ -52,55 +55,105 @@ export const settings: BlockSettings = {
...
+            label: 'Show Background',
+            defaultValue: false,
+            show: (bundle: Bundle): boolean => bundle.getBlock(PREVIEW_MODE)?.value === BlockPreview.Image,
+        },
         {
             id: HAS_BORDER_ID,
             type: 'switch',
-            label: 'Show border',
             defaultValue: false,
+            label: 'Show border',
```
Walkeezy, line 69: You could use the `getBorderSettings` from the shared package for the border settings.

PR #431, opened 2022-11-30
refactor: use new `guideline-blocks-settings` package

SamuelAlev:
Remove `block-settings` as it is now in https://github.com/Frontify/frontify-cli/tree/main/packages/guideline-blocks-settings
Use the new `defineBlock` and `defineSettings` methods
Update CLI and guideline-blocks-settings


triggertoo (approved):
💯

PR #436, opened 2022-12-02
fix: setBlockSettings only with saved changes

Walkeezy:
Removes the `...blockSettings` spread on the following blocks, because only changed settings should be saved. Otherwise it's going to be an issue on translated pages.

- Personal Note
- Sketchfab
- Storybook
- Text
- Do & Dont
- Color scale
- Checklist


celine-s (commented):
Is there a reason that you didn't change it within the `ChecklistBlock` or can it there also be removed?

celine-s (approved):

mike-schmid (approved):

PR #437, opened 2022-12-06
fix: button chooser and paddings

mike-schmid:


packages/callout-block/src/types.ts
```diff
@@ -43,9 +43,9 @@ export enum Padding {
 }
 
 export const paddingMap: Record<Padding, string> = {
-    [Padding.S]: 'tw-px-8 tw-py-4',
-    [Padding.M]: 'tw-px-8 tw-py-6',
-    [Padding.L]: 'tw-p-9',
+    [Padding.S]: 'tw-px-4 tw-py-3',
+    [Padding.M]: 'tw-px-[25px] tw-py-5',
+    [Padding.L]: 'tw-px-[30px] tw-py-[25px]',
```
mike-schmid, line 48: The new paddings are due to a smoother migration from the old block to the new one. (the "s" size looks more similar). The new numbers come from Martin and are Design approved :)

Walkeezy (approved):

triggertoo (approved):

PR #439, opened 2022-12-08
feat(TextBlock): adapt column behaviour and add column Break button

celine-s:


packages/text-block/src/TextBlock.spec.ct.tsx
```diff
@@ -22,32 +22,18 @@ describe('Text Block', () => {
         cy.get(RichTextEditor).find('[contenteditable=true]').should('not.exist');
     });
 
-    it('should be able input to two text blocks when in edit mode', () => {
```
triggertoo, line 25: not needed anymore?

packages/text-block/src/getPlugins.ts
```diff
@@ -0,0 +1,11 @@
...
+
+import { defaultPlugins, defaultPluginsWithColumns } from '@frontify/fondue';
+
+export const getPlugins = (columnNumber: number) => {
+    if (columnNumber > 1) {
+        return defaultPluginsWithColumns;
+    } else {
+        return defaultPlugins;
+    }
+};
```
triggertoo, line 11:
```suggestion
export const getPlugins = (columnNumber: number) => columnNumber > 1 ? defaultPluginsWithColumns : defaultPlugins;
```

celine-s: oh true! Removed file and inlined the condition there 👍

packages/text-block/src/TextBlock.spec.ct.tsx
celine-s: not really, as the behaviour of two or more columns is now just solved with tailwind classes. And this is tested in the last test. Do you think it needs to be tested more detailed?

mike-schmid (approved):

packages/text-block/src/TextBlock.tsx
```diff
@@ -18,40 +17,30 @@ export const TextBlock: FC<BlockProps> = ({ appBridge }) => {
...
-
+    const onTextChange = (value: string) => setBlockSettings({ content: value });
     return (
         <div
             data-test-id="text-block"
             style={{
                 gap: isColumnGutterCustom ? columnGutterCustom : columnGutterSimple,
             }}
-            className={`text-block tw-grid ${GRID_CLASSES[columnNumber]}`}
+            className={`tw-gap-2 tw-block ${hasRichTextValue(content) && GRID_CLASSES[columnNumber]}`}
```
Walkeezy, line 33:
```suggestion
            className={`tw-gap-2 tw-block ${hasRichTextValue(content) ? GRID_CLASSES[columnNumber]  : ''}`}
```
otherwise it outputs `false` as a className

Walkeezy (approved):

PR #445, opened 2022-12-20
feat(DoDontBlock): update do dont block

fulopdaniel:
Do & Don't V2.

Figma spec:
https://www.figma.com/file/pYt8ECuB6Sc63cVtpKiDgf/Blocks---Updated-Q4-2022?node-id=1097%3A266625&t=sbLqYEdhwDImFeo3-0

### DIFF
--- a/packages/dos-donts-block/src/DoDontItem.tsx
+++ b/packages/dos-donts-block/src/DoDontItem.tsx
@@ -1,89 +1,312 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
-import { IconCheckMarkCircle, IconCrossCircle, IconSize, RichTextEditor } from '@frontify/fondue';
-import { joinClassNames, toRgbaString, useGuidelineDesignTokens } from '@frontify/guideline-blocks-shared';
-import { CSSProperties, FC, useMemo } from 'react';
-import { DoDontItemProps, DoDontStyle, DoDontType, EditorChild, EditorElement } from './types';
+import { RichTextEditor, debounce } from '@frontify/fondue';
+import { joinClassNames, toRgbaString } from '@frontify/guideline-blocks-shared';
+import { CSSProperties, useEffect, useMemo, useRef, useState } from 'react';
+import autosize from 'autosize';
+import IconComponent from './components/IconComponent';
+import ItemToolbar from './components/ItemToolbar';
+import { BlockMode, DoDontItemProps, DoDontStyle, DoDontType, SortableDoDontItemProps } from './types';
+import { useSortable } from '@dnd-kit/sortable';
+import React from 'react';
+import ImageComponent from './components/ImageComponent';
+import { Asset, useAssetUpload, useBlockAssets, useFileInput } from '@frontify/app-bridge';
 
-export const DoDontItem: FC<DoDontItemProps> = ({
-    id,
-    type,
-    style,
-    doColor,
-    dontColor,
-    saveItem,
-    title = '',
-    body = '',
-    editing = false,
-}) => {
-    const doColorString = toRgbaString(doColor);
-    const dontColorString = toRgbaString(dontColor);
+export const DoDontItem = React.forwardRef<HTMLDivElement, DoDontItemProps>(
+    (
+        {
+            id,
+            type,
+            style,
+            doColor,
+            dontColor,
+            onChangeItem,
+            title = '',
+            body = '',
+            editing = false,
+            onRemoveSelf,
+            hasCustomDoIcon,
+            hasCustomDontIcon,
+            dontIconAsset,
+            doIconAsset,
+            dontIconChoice,
+            doIconChoice,
+            hasStrikethrough,
+            isDragging = false,
+            columns,
+            replaceWithPlaceholder = false,
+            transformStyle = {},
+            draggableProps = {},
+            appBridge,
+            linkedImage,
+            minRowHeight,
+            mode,
+            customImageHeightValue,
+            imageDisplay,
+            imageHeightChoice,
+            isCustomImageHeight,
+            backgroundColor,
+            borderColor,
+            borderStyle,
+            borderWidth,
+            hasBackground,
+            hasBorder,
+            hasRadius,
+            radiusChoice,
+            radiusValue,
+            designTokens,
+        },
+        ref
+    ) => {
+        const doColorString = toRgbaString(doColor);
+        const dontColorString = toRgbaString(dontColor);
+        const [isFlyoutOpen, setIsFlyoutOpen] = useState(false);
+        const { blockAssets, updateAssetIdsFromKey } = useBlockAssets(appBridge);
+        const { itemImages } = blockAssets;
+        const [internalTitle, setInternalTitle] = useState(title);
+        const titleRef = useRef<HTMLTextAreaElement>(null);
+        const [isMenuOpen, setIsMenuOpen] = useState<boolean>(false);
 
-    const headingStyles: Record<DoDontType, CSSProperties> = {
-        [DoDontType.Do]: { color: doColorString },
-        [DoDontType.Dont]: { color: dontColorString },
-    };
+        const [isUploadLoading, setIsUploadLoading] = useState(false);
+        const [openFileDialog, { selectedFiles }] = useFileInput({ multiple: false });
+        const [uploadFile, { results: uploadResults, doneAll }] = useAssetUpload({
+            onUploadProgress: () => !isUploadLoading && setIsUploadLoading(true),
+        });
 
-    const dividerStyles: Record<DoDontType, CSSProperties> = {
-        [DoDontType.Do]: { backgroundColor: doColorString },
-        [DoDontType.Dont]: { backgroundColor: dontColorString },
-    };
+        const headingColor = type === DoDontType.Do ? doColorString : dontColorString;
 
-    const isEmpty = (text: string): boolean => {
-        if (text && text !== '') {
-            return JSON.parse(text).every((element: EditorElement) => {
-                const elementsWithText = element.children.filter((child: EditorChild) => child.text !== '');
-                return elementsWithText.length === 0;
-            });
-        }
-        return true;
-    };
+        const dividerStyles: Record<DoDontType, CSSProperties> = {
+            [DoDontType.Do]: { backgroundColor: doColorString },
+            [DoDontType.Dont]: { backgroundColor: dontColorString },
+        };
 
-    const shouldBlurIcon = useMemo(() => isEmpty(title), [title]);
+        const onOpenAssetChooser = () => {
+            appBridge.openAssetChooser(
+                (result: Asset[]) => {
+                    setIsUploadLoading(true);
+                    const imageId = result[0]?.id;
+                    const existingIds = itemImages?.map((x) => x.id) || [];
+                    const newIds = [...new Set([...existingIds, imageId])];
+                    updateAssetIdsFromKey('itemImages', newIds).then(() => {
+                        onChangeItem(id, imageId, 'imageId');
+                        setIsUploadLoading(false);
+                    });
+                    appBridge.closeAssetChooser();
+                },
+                {
+                    multiSelection: false,
+                }
+            );
+        };
 
-    const { designTokens } = useGuidelineDesignTokens();
+        const onUploadClick = () => {
+            openFileDialog();
+        };
 
-    return (
-        <div>
-            <div
-                data-test-id="dos-donts-heading"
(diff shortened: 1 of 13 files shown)


packages/dos-donts-block/src/DosDontsBlock.tsx
```diff
@@ -1,95 +1,421 @@
...
+    ) => {
+        const newItems: Item[] = localItems.map((item) => (item.id === itemId ? { ...item, [type]: value } : item));
+        setAndSaveItems(newItems);
+    };
 
+    const setAndSaveItems = (newItems: Item[]) => {
+        setLocalItems(newItems);
         setBlockSettings({
-            items: [...updatedItems],
+            ...blockSettings,
```
Walkeezy, line 176: Please remove the `...blockSettings` spread here. Only save the key value pair which actually needs to be updated. Otherwise we have issues on translated pages, as only settings that differ from the main language should be saved to the i18n-settings.

packages/dos-donts-block/src/DoDontItem.tsx
```diff
@@ -1,89 +1,309 @@
...
+        };
+
+        const shouldBlurIcon = useMemo(() => isEmpty(title), [title]);
+
+        const titleRichTextEditor = useMemo(() => {
+            return (
+                <RichTextEditor
+                    designTokens={designTokens ?? undefined}
+                    value={title}
+                    onTextChange={(value) => debounce(() => onChangeItem(id, value, 'title'), 500)()}
```
Walkeezy, line 152: `onTextChange` from the RichTextEditor is already debounce, don't think it's necessary to do it here again. Also maybe it would make sense to save the item also `onBlur`, @mike-schmid just did that in the other blocks in this PR: https://github.com/Frontify/guideline-blocks/pull/446

packages/dos-donts-block/src/components/IconComponent.tsx
```diff
@@ -0,0 +1,52 @@
...
+    }, [hasCustomIcon, iconAsset]);
+
+    if (!hasCustomIcon) {
+        const Component = ICON_MAP[iconChoice];
+        return <Component />;
+    } else if (svg) {
+        return (
+            <div
+                className="tw-w-[24px] tw-h-[24px] [&>svg]:tw-w-[24px] [&>svg]:tw-h-[24px]"
+                dangerouslySetInnerHTML={{ __html: svg }}
```
Walkeezy, line 44: Maybe there is some sanitization required, see: https://www.loom.com/share/84ea3bd1967e47f88ee02f2c31a558d1

packages/dos-donts-block/src/components/ImageComponent.tsx
```diff
@@ -0,0 +1,154 @@
...
+    radiusChoice,
+    radiusValue,
+}: ImageComponentProps) => {
+    const imageHeight = useMemo(() => {
+        if (isCustomImageHeight) {
+            return customImageHeightValue;
+        } else {
+            return IMAGE_HEIGHT_VALUES[imageHeightChoice];
+        }
+    }, [customImageHeightValue, imageHeightChoice, isCustomImageHeight]);
```
Walkeezy, line 38: Why is the `useMemo` needed? I think this could simplified to the following (also for all the other `useMemo` below)
```suggestion
    const imageHeight = isCustomImageHeight ? customImageHeightValue : IMAGE_HEIGHT_VALUES[imageHeightChoice];
```

packages/dos-donts-block/src/DosDontsBlock.tsx
```diff
@@ -1,95 +1,421 @@
...
-    useEffect(() => setItems(layout === DoDontLayout.Stacked ? columns * 2 : 2), [layout, columns]);
+    const handleDragEnd = (event: DragEndEvent) => {
+        const { active, over } = event;
+        if (over && active.id !== over.id) {
+            const oldIndex = localItems.findIndex((i) => i.id === active.id);
+            const newIndex = localItems.findIndex((i) => i.id === over.id);
+            const sortedItems = arrayMove(localItems, oldIndex, newIndex);
+            setLocalItems(sortedItems);
+            setBlockSettings({
+                ...blockSettings,
```
Walkeezy, line 198: Remove `...blockSettings` here too

packages/dos-donts-block/src/DoDontItem.tsx
```diff
@@ -1,89 +1,309 @@
...
+            );
+        }, [designTokens, title, editing, onChangeItem, id]);
+
+        const descriptionRichTextEditor = useMemo(() => {
+            return (
                 <RichTextEditor
                     designTokens={designTokens ?? undefined}
                     value={body}
-                    onTextChange={(value) => saveItem(id, value, 'body')}
+                    onTextChange={(value) => debounce(() => onChangeItem(id, value, 'body'), 500)()}
```
Walkeezy, line 165: Same here with the `debounce`

packages/dos-donts-block/src/DosDontsBlock.tsx
```diff
@@ -1,95 +1,421 @@
...
+                batchAddItems(result);
+                appBridge.closeAssetChooser();
+            },
+            {
+                multiSelection: true,
+            }
+        );
+    };
+
+    const gridClassName = useMemo(() => {
```
Walkeezy, line 230: Same here with `useMemo`, is it really needed?

packages/dos-donts-block/src/components/ImageComponent.tsx
```diff
@@ -0,0 +1,154 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { Button, ButtonEmphasis, IconArrowCircleUp, IconImageStack, IconPlus32, LoadingCircle } from '@frontify/fondue';
+import { joinClassNames, radiusStyleMap, toRgbaString } from '@frontify/guideline-blocks-shared';
+import { useMemo } from 'react';
+import { DoDontImageHeight, IMAGE_HEIGHT_VALUES, ImageComponentProps, ImageFitChoice } from '../types';
+
+const ImageComponent = ({
```
triggertoo, line 8: This component contains a lot of logic. Could you split it up and provide the properties in a react context, so you can avoid to pass down all the props?

packages/dos-donts-block/src/components/ImageComponent.tsx
triggertoo: `useMemo` should only be used when you want to explicitly avoid rerendering of a component

packages/dos-donts-block/src/components/ItemToolbar.tsx
```diff
@@ -0,0 +1,155 @@
...
+    IconImageStack20,
+    IconTrashBin16,
+    IconTrashBin20,
+    MenuItemContentSize,
+} from '@frontify/fondue';
+import { joinClassNames } from '@frontify/guideline-blocks-shared';
+import React from 'react';
+import { DoDontType, ItemToolbarProps } from '../types';
+
+const ItemToolbar = React.memo(
```
triggertoo, line 19: avoid `React.memo(` here (see comment above)

packages/dos-donts-block/src/types.ts
```diff
@@ -74,3 +221,9 @@ export const columnsClasses: Record<number, string> = {
     3: 'tw-grid-cols-3',
     4: 'tw-grid-cols-4',
 };
+
+export enum DoDontVisibility {
```
triggertoo, line 225: remove?

packages/dos-donts-block/src/DosDontsBlock.tsx
```diff
@@ -1,95 +1,458 @@
...
-        } else {
-            updatedItems = updatedItems.slice(0, numberOfItems);
+    /**
+     * Create placeholders on mount if empty
+     */
+    useEffect(() => {
+        if (localItems.length === 0) {
+            const placeholderItems: Item[] = [
+                {
+                    id: cuid(),
```
SamuelAlev, line 107: You can remove the lib and use `window.crypto.randomUUID()`

packages/dos-donts-block/src/settings.ts
```diff
@@ -1,142 +1,464 @@
...
     presetCustomValue,
 } from '@frontify/guideline-blocks-settings';
-import { DoDontLayout, DoDontSpacing, DoDontStyle, spacingValues } from './types';
+import { Radius, getBorderRadiusSettings, getBorderSettings, radiusStyleMap } from '@frontify/guideline-blocks-shared';
+import {
+    BlockMode,
+    DoDontImageHeight,
+    DoDontSpacing,
+    DoDontStyle,
+    DoDontVisibility,
```
triggertoo, line 19: can be removed

packages/dos-donts-block/src/settings.ts
```diff
@@ -1,142 +1,464 @@
...
 export const DONT_COLOR_DEFAULT_VALUE = { red: 255, green: 55, blue: 90, alpha: 1 };
 
-const SPACING_VALUE_ID = 'spacingValue';
-const SPACING_CHOICE_ID = 'spacingChoice';
-
 export const settings = defineSettings({
     main: [
         {
-            id: 'style',
+            id: 'mode',
```
triggertoo, line 32: Could you comment or remove `mode`, so we don't have it for the initial rollout of the text block?

fulopdaniel: Sure, done!

packages/dos-donts-block/src/settings.ts
fulopdaniel: Done!

packages/dos-donts-block/src/DosDontsBlock.tsx
fulopdaniel: I think this would not work in the following scenario.
[1,2,3,4] are the current items. [3] is deleted, and then I add a new item, the new ID list would be [1,2,4,4].

However I removed CUID and reused the `generateRandomId` from fondue.

packages/dos-donts-block/src/types.ts
fulopdaniel: Yep!

packages/dos-donts-block/src/components/ItemToolbar.tsx
fulopdaniel: Yep, done

packages/dos-donts-block/src/components/ImageComponent.tsx
fulopdaniel: Hey, yeah I do realise there is some logic here, but I think it might still be acceptable in this case. The whole file is 100 lines, and it either shows the image or shows the button. How would you split it up? I'd be happy to do it if you have a nice suggestion.

Providing the props in react context would not reduce the lines too much, because I would still need to destructure it that would bloat the component either way, so I might prefer to keep it consistent to other components and keep the prop drilling.

packages/dos-donts-block/src/DosDontsBlock.tsx
fulopdaniel: Done, removed

packages/dos-donts-block/src/settings.ts
```diff
@@ -1,142 +1,450 @@
...
+                    defaultValue: DONT_COLOR_DEFAULT_VALUE,
+                },
+            ],
         },
         {
-            id: 'dontColor',
-            type: 'colorInput',
-            label: "Don't color",
-            defaultValue: DONT_COLOR_DEFAULT_VALUE,
+            id: 'styleImagesSection',
```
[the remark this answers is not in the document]
fulopdaniel: That is handled automatically, if none of the children are rendered.

packages/dos-donts-block/src/components/ImageComponent.tsx
triggertoo: i get your point, let's keep it as props. as an improvement and to reduce the number of props, you could i.e define `border` as a prop instead [of having `borderColor`, `borderStyle` and `borderWidth`](https://github.com/Frontify/guideline-blocks/pull/445/files#diff-2de7de30c652f8231b1363e589b960fcf2387e09d4c49c316aeddd1f78a0bbf1R23-R25). wdyt?

Walkeezy (approved):
Block looks good now from my side, nice work Daniel! 🎉

packages/dos-donts-block/src/components/ImageComponent.tsx
fulopdaniel: Yeah, good idea! Updated the PR.

triggertoo (approved):

PR #449, opened 2023-01-18
feat(ColorScaleBlock): adapt resizing and drag and drop

celine-s:


dominiquemb (changes requested):
@celine-s I tested this locally and noticed a few things need attention.

1) The grey placeholders are not rendering when the block is empty: 
Screenshot 2023-01-19 at 13 44 09


2) When I first add a color, the block appears empty with a border around it:
Screenshot 2023-01-19 at 13 44 18

3) When I add a 2nd color, the 1st color renders, however it only takes up the minimum width. All colors, unless they have been resized manually, should always take up an even amount of space in the block. If there are 2 colors, they should each have 50% width, and if there are 3, they should take up 33% width, and so on.
Screenshot 2023-01-19 at 13 44 31

4) Despite the fact that I did not resize the 2nd color, when I add a 3rd color, the 2nd color seems to take up an unexpected amount of space.
Screenshot 2023-01-19 at 13 45 16

celine-s:
> 

Uh, good catch. Should be fixed. Can you check again?

dominiquemb (changes requested):
@celine-s A few issues still remain, but it mostly looks good.

1) The delete button should be the same color as the settings buttons on the right of the block.
Screenshot 2023-01-19 at 15 59 08

2. When dragging, the color square has a white line at the bottom which shouldn't be there.
Screenshot 2023-01-19 at 16 01 34

3. When there is no room to add a new color, the 'Add Color' button should be disabled. It should also be re-enabled whenever there is extra room for a new color.
Screenshot 2023-01-19 at 16 00 45

4. When hovering over the resize handle, the cursor should change to indicate that dragging is possible.
Screenshot 2023-01-19 at 16 01 26

celine-s:
@dominiquemb 

Thanks for checking again!!

1. nice catch. Fixed.
2. True that. I did not find a fast fix therefore I added it to the ticket [here](https://app.clickup.com/t/86775v2qg) that it is still remaining and we can tackle it later.
3. At the moment it will always take from all the colors percentage wise space. As long as this is possible it would be possible to add colors. Is the wanted behaviour to just check the last color?
4. I can not reproduce it. I moved the handle a bit to the left maybe that was the issue. Or do you still have it?
5. shouldn't be the case. I think you have some old colors in it. Could you delete the block and add a new one. Do you then still have it?

dominiquemb:
> @dominiquemb Thanks for checking again!!
> 
> According to your points
> 
> 1. nice catch. Fixed.
> 2. True that. I did not find a fast fix therefore I added it to the ticket [here](https://app.clickup.com/t/86775v2qg) that it is still remaining and we can tackle it later.
> 3. At the moment it will always take from all the colors percentage wise space. As long as this is possible it would be possible to add colors. Is the wanted behaviour to just check the last color?
> 4. I can not reproduce it. I moved the handle a bit to the left maybe that was the issue. Or do you still have it?
> 5. shouldn't be the case. I think you have some old colors in it. Could you delete the block and add a new one. Do you then still have it?

3. When a color has been resized, its width should not be adjusted automatically whenever a new color is added. As a result, sometimes there may be one big color, and one small color at the end, resulting in there not being enough space to add any new colors. In this case, the 'add color' button needs to be disabled.  I can describe the desired behavior to you on a call, if you want.

4. When I hover over the drag handle, I still cannot see a 'resize' cursor. 
5. Same thing, I still cannot see the grey placeholders yet.

packages/color-scale-block/src/ColorScaleBlock.tsx
```diff
@@ -99,368 +79,152 @@ export const ColorScaleBlock: FC<BlockProps> = ({ appBridge }) => {
...
+                                                    };
+                                                    window.addEventListener('mousemove', onMouseMove);
+                                                    window.addEventListener('mouseup', onMouseUp);
+                                                }}
+                                            >
+                                                <div
+                                                    data-test-id="drag-handle"
+                                                    className="tw-bg-black-20 tw-rounded-full tw-w-2 tw-h-6"
+                                                />
+                                            </div>
```
celine-s, line 219: These lines are more or less the key of this open PR. For someone who takes it over: this resizing function is straight forward and nice to have for this block. 

cc @ragi96

PR #450, opened 2023-01-19
feat(shared): adjust useGuidelineDesignTokens to new api endpoint

Walkeezy:
Adjust `useGuidelineDesignTokens` hook to use https://github.com/Frontify/clarify/pull/13018

Issue: new endpoint does not return all (default) values

### DIFF
--- a/packages/quote-block/src/QuoteBlock.tsx
+++ b/packages/quote-block/src/QuoteBlock.tsx
@@ -35,8 +35,6 @@ export const QuoteBlock: FC<BlockProps> = ({ appBridge }) => {
     const { designTokens } = useGuidelineDesignTokens();
     const { blockAssets } = useBlockAssets(appBridge);
 
-    console.log({ designTokens });
-
     const isQuotationMarkType = blockSettings.type !== QuoteType.Indentation;
     const isFullWidth =
         blockSettings.quotationMarksAnchoring !== QuotationMarksAnchoring.HugText && isQuotationMarkType;
--- a/packages/quote-block/src/QuoteBlock.tsx
+++ b/packages/quote-block/src/QuoteBlock.tsx
@@ -35,6 +35,8 @@ export const QuoteBlock: FC<BlockProps> = ({ appBridge }) => {
     const { designTokens } = useGuidelineDesignTokens();
     const { blockAssets } = useBlockAssets(appBridge);
 
+    console.log({ designTokens });
+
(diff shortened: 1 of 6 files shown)


packages/quote-block/src/QuoteBlock.tsx
```diff
@@ -35,6 +35,8 @@ export const QuoteBlock: FC<BlockProps> = ({ appBridge }) => {
     const { designTokens } = useGuidelineDesignTokens();
     const { blockAssets } = useBlockAssets(appBridge);
 
+    console.log({ designTokens });
```
triggertoo, line 38: remove ;)

packages/shared/src/helpers/mapToGuidelineDesignTokens.ts
```diff
@@ -22,13 +22,16 @@ const TokenNameMapper: Record<string, DesignTokenName> = {
 const transformDesignTokens = (dataToTransform: DesignTokenProperties) => {
     const cssStyles: TokenValues = {};
 
-    for (const [key, value] of Object.entries(dataToTransform)) {
-        if (typeof value === 'object') {
-            transformObjectValues(key, cssStyles, value);
-        } else {
-            transformStringValues(key, cssStyles, value);
+    if (dataToTransform) {
```
triggertoo, line 25: you could set a default value for the parameter instead

packages/shared/src/helpers/mapToGuidelineDesignTokens.ts
```diff
@@ -22,13 +22,16 @@ const TokenNameMapper: Record<string, DesignTokenName> = {
...
+        for (const [key, value] of Object.entries(dataToTransform)) {
+            if (typeof value === 'object') {
+                transformObjectValues(key, cssStyles, value);
```
triggertoo, line 28: it would be an improvement if these transform functions would return instead of mutating an object :)

packages/quote-block/src/QuoteBlock.tsx
Walkeezy: Removed

packages/shared/src/helpers/mapToGuidelineDesignTokens.ts
Walkeezy: Not possible to have a default value, because it could be any css property. Changed it to return `null` now.

Walkeezy:
Change might not be needed.

Walkeezy:
Not needed anymore.

PR #455, opened 2023-01-31
fix(shared): check for `null` values

Walkeezy:
Also revert Fondue in Callout and Quote block, because the RTE is broken in the newest version.

- Callout block -> already uploaded a new version of the block as a hotfix (v17)
- Quote block -> need to upload a new version to fix this: https://app.clickup.com/t/2523021/TASK-5685

### DIFF
--- a/packages/shared/src/helpers/mapToGuidelineDesignTokens.ts
+++ b/packages/shared/src/helpers/mapToGuidelineDesignTokens.ts
@@ -122,7 +122,10 @@ export const mapToGuidelineDesignTokens = (dataToTransform: Partial<Record<strin
 
     for (const [key, value] of Object.entries(enrichedDataToTransform)) {
         const designTokenName = TokenNameMapper[key] ?? key;
-        transformedDesignTokens[designTokenName] = transformDesignTokens(value) as TokenValues;
+        if (value) {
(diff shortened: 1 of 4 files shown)


packages/shared/src/helpers/mapToGuidelineDesignTokens.ts
```diff
@@ -122,7 +122,10 @@ export const mapToGuidelineDesignTokens = (dataToTransform: Partial<Record<strin
 
     for (const [key, value] of Object.entries(enrichedDataToTransform)) {
         const designTokenName = TokenNameMapper[key] ?? key;
-        transformedDesignTokens[designTokenName] = transformDesignTokens(value) as TokenValues;
+        if (value) {
```
silviojaeger, line 125: Should we maybe explicitly check with:

```suggestion
        if (value !== null) {
```

Else we will skip all falsy values like empty strings, number 0 etc. Or is this never the case here?

Walkeezy: Good point, thanks! 👍🏻

silviojaeger (approved):
🚀

triggertoo (approved):

PR #456, opened 2023-02-03
feat(ImageBlock): add settings

Walkeezy:
This PR adds the settings for the Image block. Not possible yet: having to the link to the global settings in the Security section. Will be done with this ticket: https://app.clickup.com/t/1web3g5

Ticket: https://app.clickup.com/t/867767kmg
Figma: https://www.figma.com/file/pYt8ECuB6Sc63cVtpKiDgf/Blocks---Updated?node-id=485%3A128059&t=4tnhTTnyqAa0X6n9-0

Also: upgrade `app-bridge` & `guideline-block-settings` in all blocks because of type conflicts.


packages/image-block/src/settings.ts
```diff
@@ -1,28 +1,299 @@
...
 export const settings = defineSettings({
     basics: [
         {
-            id: 'image',
-            type: 'assetInput',
-            size: AssetInputSize.Small,
-            objectTypes: [AssetChooserObjectType.ImageVideo],
+            id: 'imageSection',
+            type: 'sectionHeading',
+            label: 'Image',
```
celine-s, line 40: For me the label in the design is the same as for the link. Therefore, I would move the label to the id `image` on line 44 and remove it here. Wdyt?

packages/image-block/src/settings.ts
```diff
@@ -1,28 +1,299 @@
...
+                },
+            ],
+        },
+    ],
+    security: [
+        {
+            id: 'security',
+            type: 'slider',
+            defaultValue: ImageSecurity.Global,
+            info: 'This setting is only available for images that are uploaded to Frontify.',
```
celine-s, line 286: In the design there is a comment `Change global settings here.`. Do you know about this?

The info `This setting is only available for images that are uploaded to Frontify.` is not shown. Probably because there is no label?

celine-s (approved):
just one small comments. Rest looks perfect!

packages/image-block/src/settings.ts
Walkeezy: Changed 👍🏻

fulopdaniel (approved):

PR #466, opened 2023-02-13
ref(dosDontsBlock): reuse BlockInjectButton

celine-s:


packages/shared/src/components/AddImagesButton.tsx
```diff
@@ -2,25 +2,31 @@
...
     isLoading?: boolean;
     label: string;
     secondaryLabel?: string;
     icon?: JSX.Element;
-    onDrop: (files: FileList) => void;
+    onDrop?: (files: FileList) => void;
     fillParentContainer?: boolean;
     onUploadClick: () => void;
     onAssetChooseClick: () => void;
+    setIsMenuOpen?: (isOpen: boolean) => void;
```
fulopdaniel, line 23: maybe rename it to something more generic? `onOpenChange` ?

packages/dos-donts-block/src/components/AddItemButton.tsx
```diff
@@ -0,0 +1,23 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { joinClassNames, textAndBgColor, textAndBgColorHover } from '@frontify/guideline-blocks-shared';
+import { AddItemButtonProps } from '../types';
+
+const AddItemButton = ({ onClick, label, icon, secondaryLabel }: AddItemButtonProps) => (
```
fulopdaniel, line 6: I think you can reuse the same component here (AddImagesButton, or if you rename it BlockInjectButton)

celine-s: I was somehow confused to reuse the other one as there is much more logic in it then in this one. It is just most of the styles it reuses. You know what I mean?

packages/shared/src/components/AddImagesButton.tsx
celine-s: true. I removed it an used the onClick which is anyway needed for the ones where the openMenu is not necessary.

packages/dos-donts-block/src/components/AddItemButton.tsx
celine-s: You are right. Removed it and made the other block more generic.

PR #472, opened 2023-02-15
feat: Add new Easing Curve block

imoutaharik:


packages/animation-curve-block/src/components/BlankSlate.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+            <CardText
+                appBridge={appBridge}
+                title={title}
+                description={description}
+                hasBorder={hasBorder}
+                designTokens={designTokens}
+                setDescription={setDescription}
+                setTitle={setTitle}
+                updateValueOnChange={true}
```
fulopdaniel, line 93: Can we save some API requests and just save on blur all the time? That way you can also omit this prop and just set it to false in CardText internally.

packages/animation-curve-block/src/components/AnimationCurveFlyout.tsx
```diff
@@ -0,0 +1,228 @@
...
+                                ...animationCurve.animationFunction,
+                                type: id as AnimationCurveType,
+                                parameters:
+                                    id === AnimationCurveType.Custom
+                                        ? { ...animationCurve.animationFunction.parameters }
+                                        : defaultAnimationCurveTypeValues[id as AnimationCurveType],
+                            },
+                        })
+                    }
+                    menuBlocks={DROPDOWN_MENU_ITEMS}
```
fulopdaniel, line 142: Choosing presets from the dropdown does not update the animation curve in the flyout. I think it would be awesome if it did, because right now you can only see the selected curve after clicking close.

packages/animation-curve-block/src/components/AnimationCurveFlyout.tsx
```diff
@@ -0,0 +1,228 @@
...
+            if (localAnimationFunction.parameters.x2 < 0 || localAnimationFunction.parameters.x2 > 1) {
+                setX2Validation(Validation.Error);
+            } else {
+                setX2Validation(Validation.Default);
+            }
+        }
+    }, [localAnimationFunction]);
+
+    const handleSave = () => {
+        onSave(animationCurve.id, { ...animationCurve });
```
fulopdaniel, line 63: If previous animation curve has title, the newly created one will copy it. I think this is a bug.

packages/animation-curve-block/src/components/AnimationCurveFlyout.tsx
```diff
@@ -0,0 +1,228 @@
...
+                            <TextInput
+                                onChange={(value) => updateAnimationCurveParameters({ y2: Number(value) })}
+                                type={TextInputType.Number}
+                                value={localAnimationFunction.parameters.y2.toString()}
+                                required
+                            />
+                        </div>
+                    </div>
+                    <div className="tw-w-20 tw-relative">
+                        <TextInput
```
fulopdaniel, line 212: Cannot add values like `0.05`

fulopdaniel:
Drag & Drop is not working for me. It is lagging - I think you should also memoize the RTEs - , and also the drag overlay is off and it is not sorting.

fulopdaniel:
Clicking the delete button on an item removes all of the items for me.

packages/animation-curve-block/src/components/BlankSlate.tsx
```diff
@@ -0,0 +1,98 @@
...
+                onSave={addAnimationCurve}
+                onCancel={() => setIsFlyoutOpen(false)}
+                onOpenChange={(isOpen) => setIsFlyoutOpen(isOpen)}
+                onAnimationCurveChange={setAnimationCurve}
+                isFlyoutOpen={isFlyoutOpen}
+                lineColor={lineColor}
+                endpointColor={endpointColor}
+                gridColor={gridColor}
+            />
+            <button
```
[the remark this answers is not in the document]
ragi96: Can't use it because of the height.

There is a problem with the layout i didn't mention, within figma the slate button and the animation curve is always 200px height, but this won't work as you are stretching the curve and so you are also not displaying the "real" curve, so I decided to it a bit different and display the slate always in the height of the other animation curves so it looks great.

packages/animation-curve-block/src/components/CardText.tsx
```diff
@@ -0,0 +1,44 @@
...
+    hasBorder,
+    designTokens,
+    updateValueOnChange,
+    isEditing,
+    setTitle,
+    setDescription,
+}: CardTextProps) => {
+    return (
+        <div className={joinClassNames([hasBorder && 'tw-px-4', 'tw-pt-4 tw-flex-1'])}>
+            <RichTextEditor
```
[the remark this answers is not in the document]
ragi96: As its within the [Figma](https://www.figma.com/file/sCEfTYxZNNKlhrpW7cCoct/New-Content-Block---Animation-curve?node-id=2133-433127&t=v8WJ8blA56sERIOK-0) i assume it should be a RTE.

And it also doesn't break the layout as its just displayed below (no alignments and so on, not like in the do & dont), i think we should discuss that with a designer...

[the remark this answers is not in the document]
fulopdaniel: If the h3 design token has margin-top I think it would add the extra space to top. Behaviours like this were considered bugs for the Do & Don't.

e.g.
https://frontify.slack.com/archives/CCXT7AFD0/p1681807431078459

packages/animation-curve-block/src/components/BlankSlate.tsx
fulopdaniel: Yep I know that. I was just wondering what does it do because in Card.tsx it is set to false explicitly, but here it is set to true explicitly.

packages/animation-curve-block/src/components/CardText.tsx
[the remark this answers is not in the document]
ragi96: Yeah but that was a bug because it wasn't aligned with the icon anymore, in this case its always below and i wouldn't consider it as bug as it just add more spacing (like defined)

packages/animation-curve-block/src/components/AnimationCurveFlyout.tsx
ragi96: Fixed that bug

packages/animation-curve-block/src/components/AnimationCurveFlyout.tsx
ragi96: fixed

packages/animation-curve-block/src/components/BlankSlate.tsx
ragi96: Its needed for clearing the `BlankSlate -> CardText -> RTE's`, after adding a new animation curve, otherwise the `setTitle('')` or `setDescription('')` won't have any effects

packages/animation-curve-block/src/components/BlankSlate.tsx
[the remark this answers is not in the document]
fulopdaniel: Yep that sounds good. But BlockInjectButton has a prop called `fillParentContainer` so you can customise its dimensions.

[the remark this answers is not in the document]
ragi96: yep, replaced it with the BlockInjectButton

fulopdaniel:
🚀

packages/animation-curve-block/src/components/AnimationCurveFlyout.tsx
```diff
@@ -0,0 +1,235 @@
...
+    const [localAnimationFunction, setLocalAnimationFunction] = useState(animationCurve.animationFunction);
+
+    useEffect(() => {
+        if (localAnimationFunction.parameters.x1) {
+            if (localAnimationFunction.parameters.x1 < 0 || localAnimationFunction.parameters.x1 > 1) {
+                setX1Validation(Validation.Error);
+            } else {
+                setX1Validation(Validation.Default);
+            }
+        }
```
triggertoo, line 53:
```suggestion
        const x1 = localAnimationFunction.parameters.x1;
        setX1Validation(x1 && (x1 >= 0 || x1 <= 1) ? Validation.Default : Validation.Error);
```

same for check below and maybe put into a function since the check is the same

packages/animation-curve-block/src/components/AnimationCurveFlyout.tsx
```diff
@@ -0,0 +1,235 @@
...
+
+    const updateLocalCurve = (animationCurvePatch: AnimationCurve): void => {
+        animationCurvePatch.animationFunction.parameters.x1 =
+            Math.round(animationCurvePatch.animationFunction.parameters.x1 * 100) / 100;
+        animationCurvePatch.animationFunction.parameters.x2 =
+            Math.round(animationCurvePatch.animationFunction.parameters.x2 * 100) / 100;
+        animationCurvePatch.animationFunction.parameters.y1 =
+            Math.round(animationCurvePatch.animationFunction.parameters.y1 * 100) / 100;
+        animationCurvePatch.animationFunction.parameters.y2 =
+            Math.round(animationCurvePatch.animationFunction.parameters.y2 * 100) / 100;
```
triggertoo, line 75: could you make this immutable / a new object?

packages/animation-curve-block/src/components/AnimationCurveFlyout.tsx
```diff
@@ -0,0 +1,235 @@
...
+                        </div>
+                        <IconCaretRight16 />
+                        <div className="tw-relative tw-w-[74px]">
+                            <span className="tw-absolute tw-top-2.5 tw-left-[3px] tw-z-10 tw-text-s tw-text-text-weak">
+                                X
+                            </span>
+                            <TextInput
+                                onChange={(value) => updateAnimationCurveParameters({ x2: Number(value) })}
+                                type={TextInputType.Number}
+                                value={localAnimationFunction.parameters.x2 as unknown as string}
```
triggertoo, line 201:
```suggestion
                                value={localAnimationFunction.parameters.x2.toString()}
```

packages/animation-curve-block/src/components/AnimationCurveFlyout.tsx
```diff
@@ -0,0 +1,235 @@
...
+                                validation={x2Validation}
+                                required
+                            />
+                        </div>
+                        <div className="tw-relative tw-w-[74px]">
+                            <span className="tw-absolute tw-top-2.5 tw-left-[3px] tw-z-10 tw-text-s tw-text-text-weak">
+                                Y
+                            </span>
+                            <TextInput
+                                onChange={(value) => updateAnimationCurveParameters({ y2: Number(value) })}
```
triggertoo, line 211: `Number(value)` can potentially return `NaN`

packages/animation-curve-block/src/components/AnimationCurveFlyout.tsx
```diff
@@ -0,0 +1,235 @@
...
+                                type={TextInputType.Number}
+                                value={localAnimationFunction.parameters.y2 as unknown as string}
```
triggertoo, line 213:
```suggestion
                                value={localAnimationFunction.parameters.y2.toString()}
```

packages/animation-curve-block/src/components/AnimationCurveFlyout.tsx
```diff
@@ -0,0 +1,235 @@
...
+                                required
+                            />
+                        </div>
+                    </div>
+                    <div className="tw-w-20 tw-relative">
+                        <TextInput
+                            onChange={(value) =>
+                                updateAnimationFunction({
+                                    ...animationCurve.animationFunction,
+                                    duration: Number(value),
```
triggertoo, line 223: catch `NaN` case

packages/animation-curve-block/src/components/AnimationCurveFlyout.tsx
triggertoo: also please use a function to not make it repetitive

packages/animation-curve-block/src/components/BlankSlate.tsx
```diff
@@ -0,0 +1,102 @@
...
+    designTokens,
+    hasBorder,
+    lineColor,
+    endpointColor,
+    gridColor,
+    canvasHeight,
+    setLocalItems,
+    setBlockSettings,
+}: BlankSlateProps) => {
+    const height = canvasHeight !== 0 ? `${canvasHeight}px` : '82px';
```
triggertoo, line 23: where do this `82px` come from? could you put it as a const with a quick description?

packages/animation-curve-block/src/components/BlankSlate.tsx
```diff
@@ -0,0 +1,102 @@
...
+    const [isFlyoutOpen, setIsFlyoutOpen] = useState(false);
+
+    const defaultAnimationCurve: AnimationCurve = {
+        id: generateRandomId(),
+        title: '',
+        description: '',
+        animationFunction: {
+            type: AnimationCurveType.EaseInOut,
+            parameters: defaultAnimationCurveTypeValues[AnimationCurveType.EaseInOut],
+            duration: 1.0,
```
triggertoo, line 35: you could also put this as a const

packages/animation-curve-block/src/components/AnimationCurveCanvasGrid.tsx
```diff
@@ -0,0 +1,37 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { Line } from './';
+import { AnimationCurveCanvasGridProps } from '../types';
+
+export const AnimationCurveCanvasGrid = ({ viewBox, lineColor }: AnimationCurveCanvasGridProps) => {
+    const backgroundGrid = [...Array(5)].map((item, i) => (
```
triggertoo, line 7: imho there's no variable need for this

packages/animation-curve-block/src/components/BlankSlate.tsx
```diff
@@ -0,0 +1,102 @@
...
+        },
+    };
+
+    const [animationCurve, setAnimationCurve] = useState<AnimationCurve>(defaultAnimationCurve);
+    const addAnimationCurve = async () => {
+        const newId = generateRandomId();
+        setLocalItems([...(content ?? []), { ...animationCurve, id: newId }]);
```
triggertoo, line 42: please use a variable since it's the same as for the settings

packages/animation-curve-block/src/components/BlankSlate.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+    useEffect(() => {
+        setAnimationCurve({ ...animationCurve, title });
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [title]);
+
+    useEffect(() => {
+        setAnimationCurve({ ...animationCurve, description });
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [description]);
```
triggertoo, line 60: you can do bot in the same `useEffect`

packages/animation-curve-block/src/components/BlankSlate.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+            <CardText
+                appBridge={appBridge}
+                title={title}
+                description={description}
+                hasBorder={hasBorder}
+                designTokens={designTokens}
+                setDescription={setDescription}
+                setTitle={setTitle}
+                updateValueOnChange={true}
```
triggertoo, line 97:
```suggestion
                updateValueOnChange
```

packages/animation-curve-block/src/components/Card.tsx
```diff
@@ -0,0 +1,238 @@
...
+                            <AnimationCanvas
+                                animationFunction={animationFunction}
+                                showGrid={hasGrid}
+                                showEndPoints={hasEndpoints}
+                                lineColor={lineColor}
+                                endpointColor={endpointsColor}
+                                gridColor={gridColor}
+                                viewBox={{
+                                    width: 350,
+                                    height: 200,
```
triggertoo, line 157: define in a const?

packages/animation-curve-block/src/components/CardText.tsx
```diff
@@ -0,0 +1,61 @@
...
+                value={description || convertToRteValue(TextStyles.ELEMENT_PARAGRAPH)}
+                placeholder="Add a description"
+                onBlur={setDescription}
+                designTokens={designTokens}
+                isEditing={isEditing}
+                updateValueOnChange={updateValueOnChange}
+                plugins={getDefaultPluginsWithLinkChooser(appBridge)}
+            />
+        );
+    }, [description, designTokens, isEditing, updateValueOnChange, setDescription, appBridge]);
```
triggertoo, line 53: isn't that basically all the properties, which would make useMemo obsolete in this case?

packages/animation-curve-block/src/constants.ts
```diff
@@ -0,0 +1,158 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { MenuBlock } from '@frontify/fondue';
+import { AnimationCurveType } from './types';
+
+export const DROPDOWN_MENU_ITEMS: MenuBlock[] = [
+    {
+        id: 'Standard',
```
triggertoo, line 8: shouldn't the id's be camelcase?

packages/animation-curve-block/src/components/AnimationCurveFlyout.tsx
ragi96: refactored the whole handling here, moved it into helper with proper tests

packages/animation-curve-block/src/components/AnimationCurveFlyout.tsx
ragi96: created a new function `validateXValue` which returns the Validation.

packages/animation-curve-block/src/components/AnimationCurveFlyout.tsx
ragi96: catched

packages/animation-curve-block/src/components/BlankSlate.tsx
ragi96: added a constant `BLANK_SLATE_INITIAL_HEIGHT`

packages/animation-curve-block/src/components/BlankSlate.tsx
ragi96: added a constant `DEFAULT_ANIMATION_FUNCTION`

packages/animation-curve-block/src/components/Card.tsx
ragi96: defined a constant `DEFAULT_ANIMATION_CANVAS_VIEWBOX`

packages/animation-curve-block/src/components/BlankSlate.tsx
ragi96: combined

packages/animation-curve-block/src/constants.ts
ragi96: you 💯  right, changed all ids to camelCase

packages/animation-curve-block/src/components/AnimationCurveCanvasGrid.tsx
ragi96: true removed it

packages/animation-curve-block/src/components/CardText.tsx
ragi96: Its not all properties. 

But i think you are right, do you think it would be better to memoize the whole CardText Component? It was added because of the [comment](https://github.com/Frontify/guideline-blocks/pull/472#issuecomment-1517452414) by @fulopdaniel

packages/animation-curve-block/src/components/AnimationCurveFlyout.tsx
```diff
@@ -0,0 +1,235 @@
...
+        });
+    };
+
+    return (
+        <Flyout
+            trigger={<></>}
+            isOpen={isFlyoutOpen}
+            onOpenChange={onOpenChange}
+            onCancel={onCancel}
+            legacyFooter={false}
```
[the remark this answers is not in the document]
ragi96: yeah that prop is deprecated but if I remove it, it will use the legacyFooter.

Because by default there is legacyFooter: true (otherwise that prop wouldn't not be needed) https://github.com/Frontify/fondue/blob/c8978e54e938a9ec74df0b3f128219213ef3b1f5/src/components/Flyout/Flyout.tsx#LL80C10-L80C10

I assume they "solved" it that way because of compatibility issues, but its just guessing from my side

packages/animation-curve-block/src/components/CardText.tsx
triggertoo: let's keep it that way if it's performance relevant 👍

triggertoo (approved):
👍

ragi96:
Closed in favor of https://github.com/Frontify/guideline-blocks/pull/553

PR #474, opened 2023-02-16
feat/add-compare-slider-block

urietberger:


ragi96:
Close, because of #520

PR #476, opened 2023-02-20
fix(DoDontBlock): unwanted padding

fulopdaniel:


packages/dos-donts-block/src/DoDontItem.tsx
```diff
@@ -154,15 +154,19 @@ export const DoDontItem = React.forwardRef<HTMLDivElement, DoDontItemProps>(
         }, [designTokens, body, id, editing, onChangeItem]);
 
         return (
-            <div ref={ref} style={{ ...transformStyle, ...(isMenuOpen ? { zIndex: 2 } : {}) }}>
+            <div
+                ref={ref}
+                className="hover:!tw-z-[2]"
```
celine-s, line 159: Could you also pass the z index in the className and remove it from style? like: 
`className={joinClassNames(["hover:!tw-z-[2]", isMenuOpen && "tw-z-[2]"])} style={transformStyle}`

celine-s: & Why exactly z-[2] and not z-10?

packages/dos-donts-block/tailwind.config.js
```diff
@@ -2,7 +2,7 @@
 
 module.exports = {
     presets: [require('@frontify/fondue/tailwind'), require('../../tailwind.config')],
-    content: ['src/**/*.{ts,tsx}'],
+    content: ['src/**/*.{ts,tsx}', '../shared/src/**/*.{ts,tsx}'],
```
celine-s, line 5: oh. Maybe it would make sense to add this to the readme within shared. Wdyt? Could it also be used like the one from fondue?

celine-s (approved):
looks good 👍

packages/dos-donts-block/src/DoDontItem.tsx
fulopdaniel: Not possible unfortunately, because when dragging, {...transformStyle} adds a zIndex:1, so I need to override that with zIndex:2 if I want that item to be on top.

packages/dos-donts-block/tailwind.config.js
fulopdaniel: yeah good idea! Added

Walkeezy (approved):

PR #480, opened 2023-02-22
feat(ImageBlock): add attachments

fulopdaniel:


packages/image-block/src/ImageBlock.tsx
```diff
@@ -1,21 +1,72 @@
...
+        }
+        newAssetIds.push(newAsset.id);
+        updateAssetIdsFromKey(ATTACHMENTS_SETTING_ID, newAssetIds);
+    };
+
+    const onAttachmentsSorted = (assets: Asset[]) => {
+        const newAssetIds = [];
+        for (const asset of assets) {
+            newAssetIds.push(asset.id);
+        }
```
celine-s, line 61: same as above: could use a mapper. Or you could already pass the ids to the function `onAttachmentsSorted = (ids: string[])`

celine-s (commented):
The attachment works really nice. 👍
One small design thing I think would be helpful for the user: on drag and drop the attachments having a grabbing instead of pointer cursor. 

Could you move the functions and consts `attachements` to `onBrowseAttachments` directly into the ImageComponent where we use it? This way we now just pass everything from the top although we use it only in the last component.
What I also was thinking is that we use in all function like e.g. in `onAddAttachments` only the ids and not the whole asset. Would it be easier to just pass the ids like `const onAddAttachments = (assetsIds: string[]) => {`

One last thing: could we pass a margin as well (like we had in the Image.tsx on line 63,64) such that the button is always in the image regardeless if we have a border or padding.

packages/image-block/src/ImageBlock.tsx
```diff
@@ -1,21 +1,72 @@
...
+        setBlockSettings({
+            attachments: attachmentsAfterReplace,
+        });
+        const newAssetIds = [];
+        for (const attachment of attachments) {
+            if (attachment.id !== attachmentToReplace.id) {
+                newAssetIds.push(attachment.id);
+            }
+        }
+        newAssetIds.push(newAsset.id);
```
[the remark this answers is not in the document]
fulopdaniel: The Asset[] is a special array with a custom iterator so filter is not available unfortunately

packages/image-block/src/ImageBlock.tsx
fulopdaniel: The Asset[] is a special array with a custom iterator so mapping is not available unfortunately

packages/image-block/src/ImageBlock.tsx
```diff
@@ -1,21 +1,72 @@
...
+            newAssetIds.push(asset.id);
+        }
+        setBlockSettings({
+            attachments: [...attachmentsInBlockSettings, ...newAssetIds.map((id) => ({ id }))],
+        });
+        addAssetIdsToKey(ATTACHMENTS_SETTING_ID, newAssetIds);
+    };
+    const onAttachmentDelete = (assetToDelete: Asset) => {
+        setBlockSettings({
+            attachments: attachmentsInBlockSettings.filter((asset) => asset.id !== assetToDelete.id),
```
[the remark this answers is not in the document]
fulopdaniel: fixed, thanks!

fulopdaniel:
> The attachment works really nice. 👍 One small design thing I think would be helpful for the user: on drag and drop the attachments having a grabbing instead of pointer cursor.
> 
> Could you move the functions and consts `attachements` to `onBrowseAttachments` directly into the ImageComponent where we use it? This way we now just pass everything from the top although we use it only in the last component. What I also was thinking is that we use in all function like e.g. in `onAddAttachments` only the ids and not the whole asset. Would it be easier to just pass the ids like `const onAddAttachments = (assetsIds: string[]) => {`
> 
> One last thing: could we pass a margin as well (like we had in the Image.tsx on line 63,64) such that the button is always in the image regardeless if we have a border or padding.
> 
> sorry for the wall of text - was just thinking out loud. Let me know what you think 🙃

Thanks for the nice suggestions! I changed the cursor to grab/grabbing and modified the position of the attachment so it is always inside the image.

For changing the interface to assetsIds I am a bit hesitant because while this usecase only needs the IDs, I can imagine a scenario where in other blocks we will need more information about the attachments uploaded, e.g. the type, or the size, so I think keeping the shared components as generic as possible is a good idea.

Regarding moving the logic deeper, I did it this way because logic regarding setting the block settings is usually in the top level of the component so it can act as a single source of truth and ensure no mismatch can happen as opposed to when setting the state inside the component.

What do you think?

packages/image-block/src/ImageBlock.tsx
```diff
@@ -1,21 +1,72 @@
...
     const image = blockAssets?.[IMAGE_SETTING_ID]?.[0];
+    const attachments = blockAssets?.[ATTACHMENTS_SETTING_ID] || [];
     const isEditing = useEditorState(appBridge);
+    const attachmentsInBlockSettings = blockSettings.attachments || [];
+
+    const onAddAttachments = (assets: Asset[]) => {
+        const newAssetIds = [];
+        for (const asset of assets) {
+            newAssetIds.push(asset.id);
+        }
```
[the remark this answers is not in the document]
celine-s: ahhh, crazy :D. than it makes sense.

packages/image-block/src/components/Image.tsx
```diff
@@ -42,50 +118,33 @@ export const Image = ({ image, blockSettings, isEditing }: ImageProps) => {
...
+                    src={image.genericUrl}
+                    alt={image.fileName}
                     style={{
-                        marginTop: padding,
-                        marginRight: padding,
+                        width: image.width,
+                        borderRadius: borderRadius ?? radiusValues[CornerRadius.None],
+                        backgroundColor: blockSettings.hasBackground
+                            ? toRgbaString(blockSettings.backgroundColor)
+                            : undefined,
```
celine-s, line 144: just saw. I think this also needs to be moved to line 111.

Walkeezy (approved):
Looks niiice!

packages/image-block/src/ImageBlock.tsx
```diff
@@ -1,21 +1,72 @@
...
 export const ImageBlock = ({ appBridge }: BlockProps) => {
     const [blockSettings, setBlockSettings] = useBlockSettings<Settings>(appBridge);
-    const { blockAssets } = useBlockAssets(appBridge);
+    const { blockAssets, updateAssetIdsFromKey, addAssetIdsToKey, deleteAssetIdsFromKey } = useBlockAssets(appBridge);
     const image = blockAssets?.[IMAGE_SETTING_ID]?.[0];
+    const attachments = blockAssets?.[ATTACHMENTS_SETTING_ID] || [];
     const isEditing = useEditorState(appBridge);
+    const attachmentsInBlockSettings = blockSettings.attachments || [];
+
+    const onAddAttachments = (assets: Asset[]) => {
```
triggertoo, line 22: you could use `useReducer` to extract this logic and pass down the dispatch action:
```
const reducer = (state, action) => {
    switch (action.type) {
        case 'add':
            ...
        case 'replace':
            ...
        case 'delete':
            ...
        case 'sorted':
            ...
}
```

see https://beta.reactjs.org/reference/react/useReducer

triggertoo: that way, you can passt down the dispatch action which you get from `useReducer` to avoid prop drilling

triggertoo: regarding implementation: for the settings itself, it would mean to introduce useReducedSettings in appBridge

fulopdaniel: Thanks Marco for the insight. I decided to create a useAttachments hook inside the shared folder with all the logic bundled in it, so it is easily reusable. Let me know what do you think!

PR #482, opened 2023-02-24
feat(DoDontItem): remove custom icon in sidebar

fulopdaniel:


celine-s (commented):
Settings work perfect. Could you also remove the dontIconAsset & doIconAsset logic from the component and type file?

fulopdaniel:
> Settings work perfect. Could you also remove the dontIconAsset & doIconAsset logic from the component and type file?

This is just a temporary measure, once asset handling is done this will be re-enabled. Similar to the Image mode in the main section. What do you think?

celine-s (approved):

Walkeezy (approved):

PR #485, opened 2023-02-27
feat(shared): provide default tokens in `useGuidelineDesignTokens`

Walkeezy:
Provide default values for all guideline tokens. Default values are collected from the UI when creating a new guideline.

How to test:
1. Serve a block, which uses the `useGuidelineDesignTokens` hook, locally (for examle the text block)
2. Check if the styles of headings, buttons etc. are matching the design tokens in the sidebar.
3. Check it also in a brand new guideline. You can create a new guideline on product.frontify.com.

Related ticket: https://app.clickup.com/t/8677fr17t


packages/text-block/src/TextBlock.tsx
```diff
@@ -23,7 +23,7 @@ export const TextBlock: FC<BlockProps> = ({ appBridge }) => {
     return (
         <div data-test-id="text-block">
             <RichTextEditor
-                designTokens={designTokens ?? undefined}
+                designTokens={designTokens}
```
rawi96: There are a few other blocks which maybe need this change?
- DoDontItem.tsx
- ImageCaption.tsx
- PersonalNoteBlock.tsx
- QuoteBlock.tsx

packages/shared/src/helpers/mergeDeep.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+    test('should overwrite value for same key on nested object', () => {
+        expect(mergeDeep({ a: { a: 1 } }, { a: { a: 'overwritten', b: 2 } })).toEqual({
+            a: { a: 'overwritten', b: 2 },
+        });
+    });
+
+    test('should overwrite value for same key on nested object', () => {
+        expect(
+            mergeDeep(
+                { image: 'pipapo', content: { text: 'hello!' } },
```
celine-s, line 48: pipapo 😂

rawi96 (commented):
I commented one little nit.

And btw maybe you could use the `deepMerge` method from the RTE in fondue instead of writing an own. src/components/RichTextEditor/utils/deepMerge.ts

And for the future we should look what we do with the default design tokens within the RTE. Maybe we could now remove them. 

Rest looks good. 🚀

packages/text-block/src/TextBlock.tsx
Walkeezy: Done 👍🏻

rawi96 (approved):

PR #488, opened 2023-03-01
feat(ImageBlock): add tests

fulopdaniel:


packages/image-block/src/ImageBlock.spec.ct.tsx
```diff
@@ -1,16 +1,112 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
-import { withAppBridgeBlockStubs } from '@frontify/app-bridge';
+import { AssetDummy, withAppBridgeBlockStubs } from '@frontify/app-bridge';
 import { mount } from 'cypress/react';
 import { ImageBlock } from './ImageBlock';
+import { ATTACHMENTS_SETTING_ID, IMAGE_SETTING_ID } from './settings';
 
 const ImageBlockSelector = '[data-test-id="image-block"]';
+const ImageBlockWithLinkSelector = '[data-test-id="image-block"] > a';
```
Walkeezy, line 9: Nit: don't see the point to define this as a const here, as it's only used once. IMO you could do something like this in the test, then it's also much clearer what the test actually does.

`cy.get(ImageBlockSelector).get('a').should('have.attr', 'href', 'https://frontify.com');`

fulopdaniel: I like it! 👍

Walkeezy (approved):

celine-s (commented):
looks good! Could you also add some test about the styles? Padding, border, align etc.?

fulopdaniel:
> 

Absolutely! Done

celine-s (approved):

PR #491, opened 2023-03-02
feat(imageBlock): caption

celine-s:
Ticket: https://app.clickup.com/t/8677a21bv

### DIFF
--- a/packages/brand-positioning-block/package.json
+++ b/packages/brand-positioning-block/package.json
@@ -30,7 +30,7 @@
     },
     "dependencies": {
         "@frontify/app-bridge": "3.0.0-beta.38",
-        "@frontify/fondue": "12.0.0-beta.35",
+        "@frontify/fondue": "12.0.0-beta.114",
         "@frontify/fondue-tokens": "3.1.0",
         "@frontify/guideline-blocks-settings": "0.25.10",
         "@frontify/guideline-blocks-shared": "workspace:*",
--- a/packages/brand-positioning-block/package.json
+++ b/packages/brand-positioning-block/package.json
@@ -30,7 +30,7 @@
     },
     "dependencies": {
         "@frontify/app-bridge": "3.0.0-beta.38",
-        "@frontify/fondue": "12.0.0-beta.114",
+        "@frontify/fondue": "12.0.0-beta.118",
         "@frontify/fondue-tokens": "3.1.0",
         "@frontify/guideline-blocks-settings": "0.25.10",
         "@frontify/guideline-blocks-shared": "workspace:*",
--- a/packages/image-block/package.json
+++ b/packages/image-block/package.json
@@ -21,7 +21,7 @@
         "@types/react": "17.0.50",
         "@types/react-dom": "17.0.17",
         "autoprefixer": "10.4.13",
-        "cypress": "12.7.0",
+        "cypress": "11.0.1",
         "eslint": "8.27.0",
         "eslint-plugin-notice": "0.9.10",
         "postcss": "8.4.19",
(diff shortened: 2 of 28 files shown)


packages/brand-positioning-block/package.json
```diff
@@ -30,7 +30,7 @@
     },
     "dependencies": {
         "@frontify/app-bridge": "3.0.0-beta.38",
-        "@frontify/fondue": "12.0.0-beta.35",
+        "@frontify/fondue": "12.0.0-beta.114",
```
Walkeezy, line 33: I would suggest to not update fondue in all the other blocks. With all the changes in fondue, we need to test the block if it still works as expected. This here is also quite a big step: version 35 to 114.

packages/image-block/src/components/helpers.ts
```diff
@@ -0,0 +1,68 @@
...
+    .setPlugin([...markStylesPlugins, new LinkPlugin(), new ButtonPlugin(), new CodePlugin()])
+    .setPlugin([
+        ...alignmentPlugins,
+        new UnorderedListPlugin(),
+        new CheckboxListPlugin(),
+        new OrderedListPlugin(),
+        new ResetFormattingPlugin(),
+    ]);
+
+// TODO: fix RTE toolbar width
```
Walkeezy, line 63: Remove?

fulopdaniel (approved):

packages/brand-positioning-block/package.json
celine-s: cypress has issues with different versions installed in the packages. the brand positioning block shouldn't have issues with the newest version, but I see your point, it would be nice to have different versions

packages/image-block/package.json
```diff
@@ -21,7 +21,7 @@
         "@types/react": "17.0.50",
         "@types/react-dom": "17.0.17",
         "autoprefixer": "10.4.13",
-        "cypress": "11.0.1",
+        "cypress": "12.7.0",
```
Walkeezy, line 24: Necessary?

packages/image-block/src/components/ImageCaption.tsx
```diff
@@ -20,40 +21,42 @@ export const ImageCaption = ({
...
-                    onBlur={onDescriptionChange}
-                    designTokens={designTokens}
-                    placeholder={isEditing ? 'Add a description here' : undefined}
-                />
-            </div>
+        <div className="tw-mt-3 tw-gap-1 tw-flex-1 tw-w-full">
+            {isEditing ? (
+                <>
+                    <RichTextEditor
+                        id="image-title"
```
Walkeezy, line 34: `id` should be unique even if there are multiple blocks on the same page. You can use `appBridge.getBlockId().toString()` and extend the string with `_imageTitle` or something like this.

packages/image-block/src/components/helpers.ts
```diff
@@ -0,0 +1,67 @@
...
+    }),
+];
+const alignmentPlugins = [
+    new AlignLeftPlugin(),
+    new AlignRightPlugin(),
+    new AlignCenterPlugin(),
+    new AlignJustifyPlugin(),
+];
+const markStylesPlugins = [new BoldPlugin(), new ItalicPlugin(), new UnderlinePlugin(), new StrikethroughPlugin()];
+captionPlugins
```
Walkeezy, line 52: Softbreak plugin not needed?

celine-s: added. Thanks 🙇‍♀️

packages/image-block/src/components/ImageCaption.tsx
celine-s: changed 👍

Walkeezy (approved):

PR #492, opened 2023-03-03
feat(CalloutBlock): implement html-view-mode & update fondue

Walkeezy:
- Display html in view mode of block, instead of the RTE
- Update fondue

### DIFF
--- a/packages/callout-block/src/CalloutBlock.spec.ct.tsx
+++ b/packages/callout-block/src/CalloutBlock.spec.ct.tsx
@@ -8,6 +8,7 @@ import { Alignment, Icon, Padding, Width } from './types';
 
 const CalloutBlockSelector = '[data-test-id="callout-block"]';
 const RichTextEditor = '[data-test-id="rich-text-editor"]';
+const HtmlContent = '[data-test-id="rte-content-html"]';
 const CalloutWrapper = '[data-test-id="callout-wrapper"]';
 const CalloutIcon = '[data-test-id="callout-icon"]';
 const CalloutIconCustom = '[data-test-id="callout-icon-custom"]';
@@ -26,7 +27,7 @@ describe('Callout Block', () => {
         const [TextBlockWithStubs] = withAppBridgeBlockStubs(CalloutBlock, {});
 
         mount(<TextBlockWithStubs />);
-        cy.get('[data-test-id="rte-content-html"]').should('exist');
+        cy.get(HtmlContent).should('exist');
         cy.get(RichTextEditor).should('not.exist');
     });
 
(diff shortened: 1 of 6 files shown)


packages/callout-block/src/CalloutBlock.tsx
```diff
@@ -90,15 +93,20 @@ export const CalloutBlock: FC<BlockProps> = ({ appBridge }) => {
...
-                    border={false}
-                    value={blockSettings.textValue}
-                    placeholder="Type your text here"
-                    designTokens={useCalloutColors(designTokens, textColor)}
-                />
+                {!isEditing ? (
+                    <div data-test-id="rte-content-html" dangerouslySetInnerHTML={{ __html: html }} />
+                ) : (
+                    <RichTextEditor
+                        data-test-id="rich-text-editor"
```
fulopdaniel, line 100: doesn't RTE has this data-test-id built-in?

fulopdaniel (approved):

packages/callout-block/src/CalloutBlock.tsx
Walkeezy: Holymoly, good point, thanks!

ragi96: Well should we really rely on "external" test ids for tests?

packages/callout-block/src/CalloutBlock.spec.ct.tsx
```diff
@@ -7,17 +7,27 @@ import { ICON_ASSET_ID } from './settings';
...
         mount(<CalloutBlockWithStubs />);
         cy.get(CalloutBlockSelector).should('exist');
+        cy.get(RichTextEditor).should('exist');
+    });
+
+    it('should not be able to input to a callout block when in view mode', () => {
+        const [TextBlockWithStubs] = withAppBridgeBlockStubs(CalloutBlock, {});
+
+        mount(<TextBlockWithStubs />);
+        cy.get('[data-test-id="rte-content-html"]').should('exist');
```
ragi96, line 29: best would be to put this also in a const, so every test id is declared

packages/callout-block/src/CalloutBlock.tsx
Walkeezy: The `data-test-id` prop on the component currently doesn't seem to have an effect anyway, so I would keep it as it for the moment. If we update fondue in the future and the 'external' test id changes, we will notice when the tests fail. :)

packages/callout-block/src/CalloutBlock.spec.ct.tsx
Walkeezy: Done 👍🏻

ragi96 (approved):

PR #497, opened 2023-03-07
fix(TextBlock): preset "auto" value with empty string

Walkeezy:
Preset "auto" value with empty string instead of "normal" when switch to custom. Before it was like this:

normal-gutter


celine-s (approved):
nice!

fulopdaniel (approved):

PR #499, opened 2023-03-08
feat(useGuidelineDesignTokens): add support for `family_css` property

Walkeezy:
Use `family_css` property if provided for the design tokens, because it might be different:

Bildschirm­foto 2023-03-08 um 10 37 35


packages/shared/src/helpers/mapToGuidelineDesignTokens.ts
```diff
@@ -43,11 +44,17 @@ const transformObjectValues = (key: string, cssStyles: TokenValues, value: Direc
...
+    cssStyles: TokenValues,
+    value: string,
+    mainFontFamily: string,
+    fontFamilyCss?: string
+) => {
     cssStyles.hover = cssStyles.hover || {};
     switch (key) {
         case DesignTokenPropertiesEnum.family:
-            cssStyles.fontFamily = getFont(value, mainFontFamily);
+            cssStyles.fontFamily = fontFamilyCss ?? getFont(value, mainFontFamily);
```
celine-s, line 57:
```suggestion
            cssStyles.fontFamily =  getFont(fontFamilyCss ?? value, mainFontFamily);
```

celine-s (approved):

packages/shared/src/hooks/useGuidelineDesignTokens.ts
```diff
@@ -7,6 +7,7 @@ import { mergeDeep } from '../helpers/mergeDeep';
 
 export enum DesignTokenPropertiesEnum {
     family = 'family',
```
triggertoo, line 9: can `family` be removed since it's a naming property?

triggertoo: or should we even map `family_css` to `family`?

triggertoo: as discussed, let's keep it since `family` provides `inherit` information

PR #505, opened 2023-03-14
feat(useAttachments): use asset id order instead of settings

fulopdaniel:
Updated app bridge version because I received the following error for each package during typecheck:

Argument of type 'AppBridgeBlock' is not assignable to parameter of type 'AppBridgeBlock | AppBridgeTheme'.


packages/shared/src/hooks/useAttachments.ts
```diff
@@ -1,47 +1,44 @@
...
-    const [blockSettings, setBlockSettings] = useBlockSettings(appBridge);
-    const { blockAssets, updateAssetIdsFromKey, addAssetIdsToKey, deleteAssetIdsFromKey } = useBlockAssets(appBridge);
+export const useAttachments = (appBridge: AppBridgeBlock, assetSettingsKey: string) => {
+    const { blockAssets, updateAssetIdsFromKey } = useBlockAssets(appBridge);
     const attachments = blockAssets?.[assetSettingsKey] || [];
-    const attachmentsInBlockSettings = (blockSettings[blockSettingsKey] || []) as { id: number }[];
 
-    const onAddAttachments = async (assets: Asset[]) => {
+    const onAddAttachments = async (newAssets: Asset[]) => {
         const newAssetIds = [];
```
ragi96, line 10:
```suggestion
        const existingAssetId = attachments.map((attachment) => attachment.id);
```

packages/shared/src/hooks/useAttachments.ts
```diff
@@ -1,47 +1,44 @@
...
-        for (const asset of assets) {
+
+        for (const attachment of attachments) {
+            newAssetIds.push(attachment.id);
+        }
```
ragi96, line 14: Can be removed if you change the initialisation

packages/shared/src/hooks/useAttachments.ts
```diff
@@ -1,47 +1,44 @@
...
+        for (const asset of newAssets) {
             newAssetIds.push(asset.id);
         }
```
ragi96, line 17:
```suggestion
        newAssetIds.push(...newAssets.map((asset) => asset.id));
```

packages/shared/src/hooks/useAttachments.ts
```diff
@@ -1,47 +1,44 @@
...
-        await setBlockSettings({
-            [blockSettingsKey]: attachmentsInBlockSettings.filter(({ id }) => id !== assetToDelete.id),
-        });
-        await deleteAssetIdsFromKey(assetSettingsKey, [assetToDelete.id]);
+        const newAssetIds = [];
+        for (const attachment of attachments) {
+            if (attachment.id !== assetToDelete.id) {
+                newAssetIds.push(attachment.id);
+            }
+        }
```
ragi96, line 27:
```suggestion
    const newAssetIds = attachments
        .filter((attachment) => attachment.id !== assetToDelete.id)
        .map((attachment) => attachment.id);
```

packages/shared/src/hooks/useAttachments.ts
```diff
@@ -50,20 +47,15 @@ export const useAttachments = (appBridge: AppBridgeBlock, blockSettingsKey: stri
         for (const asset of assets) {
             newAssetIds.push(asset.id);
         }
```
ragi96, line 49: could be something like that:

```suggestion
        const newAssetIds = assets.map((asset) => asset.id);
```

packages/shared/src/hooks/useAttachments.ts
```diff
@@ -1,47 +1,44 @@
...
-        });
         const newAssetIds = [];
+
         for (const attachment of attachments) {
             if (attachment.id !== attachmentToReplace.id) {
                 newAssetIds.push(attachment.id);
+            } else {
+                newAssetIds.push(newAsset.id);
             }
         }
```
ragi96, line 41: Could be something like that:

```suggestion
         const newAssetIds = attachments.map((attachment) => {
            if (attachment.id !== attachmentToReplace.id) {
                return attachment.id;
            } else {
                return newAsset.id;
            }
        });
```

ragi96 (commented):
Update also the app-bridge within the examples

packages/shared/src/hooks/useAttachments.ts
fulopdaniel: Good catch. As discussed, for when the assets array comes from the asset chooser it has a custom iterable which does not support map. For the other occasions I updated the code.

ragi96 (approved):
🚀

packages/shared/src/hooks/useAttachments.ts
```diff
@@ -1,69 +1,46 @@
...
+        await updateAssetIdsFromKey(assetSettingsKey, newAssetIds);
+    };
 
     return {
         onAddAttachments,
         onAttachmentDelete,
         onAttachmentReplace,
         onAttachmentsSorted,
-        sortedAttachments,
+        sortedAttachments: attachments,
```
SamuelAlev, line 44:
```suggestion
       attachments,
```

packages/audio-block/src/components/BlockAttachments.tsx
```diff
@@ -11,7 +11,7 @@ export const BlockAttachments = ({ appBridge }: BlockAttachmentsProps) => {
         onAttachmentDelete,
         onAttachmentReplace,
         onAttachmentsSorted,
-    } = useAttachments(appBridge, ATTACHMENTS_SETTING_ID, ATTACHMENTS_SETTING_ID);
+    } = useAttachments(appBridge, ATTACHMENTS_SETTING_ID);
```
SamuelAlev, line 14:
```suggestion
    } = useAttachments(appBridge, ATTACHMENTS_ASSET_ID);
```

It isn't a setting ID but an asset ID

packages/image-block/src/components/Image.tsx
```diff
@@ -65,7 +65,7 @@ export const ImageComponent = ({ image, appBridge, blockSettings }: ImageProps)
         onAttachmentDelete,
         onAttachmentReplace,
         onAttachmentsSorted,
-    } = useAttachments(appBridge, ATTACHMENTS_SETTING_ID, ATTACHMENTS_SETTING_ID);
+    } = useAttachments(appBridge, ATTACHMENTS_SETTING_ID);
```
SamuelAlev, line 68:
```suggestion
    } = useAttachments(appBridge, ATTACHMENTS_ASSET_ID);
```

packages/image-block/src/components/Image.tsx
SamuelAlev, line 1: You can also remove the `attachments: { id: number }[];` in `types.ts`

Walkeezy (approved):

SamuelAlev (approved):

PR #509, opened 2023-03-16
feat: audio block accessibility improvements

ragi96:
Hopefully the last round of changes for the audio block... 

Improved UX (misunderstood Martin on the part about when the outline with the toolbar should shine, outline in Edit Mode and just on hover or focus)

Improved Accessibility Toolbar is now "fully" accessible by keyboard

### DIFF
--- a/packages/audio-block/src/AudioBlock.tsx
+++ b/packages/audio-block/src/AudioBlock.tsx
@@ -47,6 +47,7 @@ import { UploadPlaceholder } from './components/UploadPlaceholder';
 import { ItemToolbar } from './components/ItemToolbar';
 import { BlockAttachments } from './components/BlockAttachments';
 import { useEffect, useMemo, useState } from 'react';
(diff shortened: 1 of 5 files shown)


packages/audio-block/src/AudioBlock.tsx
```diff
@@ -157,19 +164,12 @@ export const AudioBlock = ({ appBridge }: BlockProps) => {
         // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [doneAll, uploadResults]);
 
+    console.log(isFocused);
+
```
celine-s, line 168:
```suggestion
```

celine-s (approved):
looks good to me. Well done! 👏

packages/audio-block/src/AudioBlock.tsx
ragi96: removed

fulopdaniel (approved):

PR #514, opened 2023-03-21
chore: update app-bridge

SamuelAlev:
Update `@frontiyf/app-bridge`, add `sinon` as it is a `peerDependencies` and add other missing `peerDependencies` in `gradient-block` and `shared`


Walkeezy (approved):
Nice! 🧹

fulopdaniel (approved):

PR #519, opened 2023-03-27
feat(imageBlock): use BlockItemWrapper and convertToRteValue

celine-s:
implements: https://app.clickup.com/t/8677fcy7n

fixes:
* only image formats allowed (error mgs in drop)
* added selected Asset id
* audio block: allow capital audio formats & active audio in assetChooser
* use convertToRteValue()

### DIFF
--- a/packages/image-block/src/ImageBlock.spec.ct.tsx
+++ b/packages/image-block/src/ImageBlock.spec.ct.tsx
@@ -3,7 +3,7 @@
 import { AssetDummy, withAppBridgeBlockStubs } from '@frontify/app-bridge';
 import { mount } from 'cypress/react';
 import { ImageBlock } from './ImageBlock';
-import { ATTACHMENTS_ASSET_ID, IMAGE_SETTING_ID } from './settings';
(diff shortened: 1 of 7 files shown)


packages/image-block/src/ImageBlock.spec.ct.tsx
```diff
@@ -84,31 +85,21 @@ describe('Image Block', () => {
...
             blockSettings: {
-                name: 'Test Name',
+                name: JSON.stringify([{ type: 'imageTitle', children: [{ text: 'Test Name' }] }]),
             },
         });
         mount(<ImageBlockWithStubs />);
-        cy.get(ImageBlockSelector).should('contain.text', 'Test Name');
-    });
-
-    it('should render the file name if no title is provided', () => {
```
ragi96, line 94: Add a test where no title is setted and the asset gets uploaded and afterwards the asset title is the title, so we have that case also

celine-s: Newly we set the title after uploading and this unfortunately cannot be mockt at the moment with the appBridge asset dummy

Walkeezy (approved):

PR #525, opened 2023-03-29
feat(AssetKitBlock): use bulk download

ragi96:
FYI: special the screen reader announcement position within the button is because the most screen readers just read whats focused, so it has to be within the button, because its focused on click

Also special styling just to hide the text, but not for screen readers


packages/asset-kit-block/src/AssetKitBlock.tsx
```diff
@@ -39,24 +79,35 @@ export const AssetKitBlock = ({ appBridge }: BlockProps): ReactElement => {
...
+                        onClick={() => startDownload()}
                         style={designTokens.buttonPrimary}
                     >
                         Download package
+                        <span
+                            data-test-id="asset-kit-block-screen-reader"
+                            ref={screenReaderRef}
+                            role="status"
+                            className="tw-absolute -tw-left-[10000px] tw-top-auto tw-w-1 tw-h-1 overflow-hidden"
+                        ></span>
```
fulopdaniel, line 96:
```suggestion
                        />
```

packages/asset-kit-block/src/AssetKitBlock.tsx
```diff
@@ -39,24 +79,35 @@ export const AssetKitBlock = ({ appBridge }: BlockProps): ReactElement => {
                 <div className="tw-flex-none">
                     <button
                         data-test-id="asset-kit-block-download-button"
-                        disabled={isDownloadingAssets || isUploadingAssets || currentAssets.length <= 0}
-                        onClick={() =>
-                            generateBulkDownload(appBridge.getProjectId(), currentAssets, setIsDownloadingAssets)
+                        disabled={
+                            [BulkDownloadState.Error, BulkDownloadState.Pending, BulkDownloadState.Started].includes(
+                                status
+                            ) || currentAssets.length <= 0
```
fulopdaniel, line 85: Minor, but I think this makes it a tiny bit easier to understand

```suggestion
                            ) || currentAssets.length === 0
```

packages/asset-kit-block/src/AssetKitBlock.tsx
```diff
@@ -39,24 +79,35 @@ export const AssetKitBlock = ({ appBridge }: BlockProps): ReactElement => {
...
                         }
+                        onClick={() => startDownload()}
                         style={designTokens.buttonPrimary}
                     >
                         Download package
+                        <span
+                            data-test-id="asset-kit-block-screen-reader"
+                            ref={screenReaderRef}
+                            role="status"
+                            className="tw-absolute -tw-left-[10000px] tw-top-auto tw-w-1 tw-h-1 overflow-hidden"
```
fulopdaniel, line 95: I wonder if offsetting to left can potentially cause the page to be scrollable horizontally. What do you think of instead of offsetting you would create a 0px wide, 0px tall element with `pointer-events: none` and `visibility: hidden` ?

packages/asset-kit-block/src/AssetKitBlock.tsx
```diff
@@ -3,23 +3,63 @@
...
+    const startDownload = () => {
+        if (downloadUrlBlock) {
+            downloadAssets(downloadUrlBlock);
+        } else {
+            generateBulkDownload(currentAssets.map((asset) => asset.id));
+        }
+    };
+
+    const downloadAssets = (downloadUrl: string) => {
+        window.open(downloadUrl, '_blank');
```
fulopdaniel, line 40: Does this open a new tab that closes automatically? If yes, can we go with an approach like in `downloadAsset.ts` ?

ragi96: Yes it does, and I adjusted it 👍

packages/asset-kit-block/src/AssetKitBlock.tsx
```diff
@@ -39,24 +81,35 @@ export const AssetKitBlock = ({ appBridge }: BlockProps): ReactElement => {
...
                         }
+                        onClick={() => startDownload()}
                         style={designTokens.buttonPrimary}
                     >
                         Download package
+                        <span
+                            data-test-id="asset-kit-block-screen-reader"
+                            ref={screenReaderRef}
+                            role="status"
+                            className="tw-absolute -tw-left-[10000px] tw-top-auto tw-w-1 tw-h-1 tw-overflow-hidden"
```
fulopdaniel, line 97: Not sure if tw-overflow-hidden has any effect here, since it would control the overflow mechanics of its content, but it has no content. I think what you maybe meant to do is add overflow-hidden to one of the parent elements?

packages/asset-kit-block/src/AssetKitBlock.tsx
```diff
@@ -39,24 +81,35 @@ export const AssetKitBlock = ({ appBridge }: BlockProps): ReactElement => {
...
                         data-test-id="asset-kit-block-download-button"
-                        disabled={isDownloadingAssets || isUploadingAssets || currentAssets.length <= 0}
-                        onClick={() =>
-                            generateBulkDownload(appBridge.getProjectId(), currentAssets, setIsDownloadingAssets)
+                        disabled={
+                            [BulkDownloadState.Error, BulkDownloadState.Pending, BulkDownloadState.Started].includes(
+                                status
+                            ) || currentAssets.length === 0
                         }
+                        onClick={() => startDownload()}
```
celine-s, line 89:
```suggestion
                        onClick={startDownload}
```

celine-s (approved):
looks good to me and works nicely! 👍

fulopdaniel (approved):

PR #533, opened 2023-04-05
refactor(shared): RichTextEditor flexibility

ragi96:
Changed the handling so its more flexible and could be used everywhere 😄

### DIFF
--- a/packages/shared/src/components/RichTextEditor/types.ts
+++ b/packages/shared/src/components/RichTextEditor/types.ts
@@ -4,9 +4,7 @@ import { PluginComposer } from '@frontify/fondue';
 import { DesignTokens } from '@frontify/fondue/dist/components/RichTextEditor/types';
 
 export type RichTextEditorProps = {
-    settingsId: string;
     isEditing: boolean;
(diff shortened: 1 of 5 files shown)


celine-s (approved):
nice!

packages/shared/src/components/RichTextEditor/types.ts
```diff
@@ -4,16 +4,15 @@ import { PluginComposer } from '@frontify/fondue';
...
     isEditing: boolean;
-    setBlockSettings: (settings: Record<string, unknown>) => void;
     value?: string;
     placeholder?: string;
     designTokens?: DesignTokens;
     columns?: number;
     gap?: string;
     plugins?: PluginComposer;
     updateValueOnChange?: boolean;
+    onStore: (value: string) => void;
```
fulopdaniel, line 15: what do you think about `onBlur` instead of onStore? onStore kind of makes an assumption about how the consumer will use this prop.

ragi96: You are right, I adjusted it.

fulopdaniel (approved):

PR #538, opened 2023-04-17
fix(gradient-block): Position Uniqueness

ragi96:
[Task](https://app.clickup.com/t/8677uqnq0)

UX Review finding which was caused problems if multiple color had the same position.

### DIFF
--- a/packages/gradient-block/src/components/ColorFlyout.tsx
+++ b/packages/gradient-block/src/components/ColorFlyout.tsx
@@ -141,7 +141,7 @@ export const ColorFlyout = ({
                     <FormControl
                         helper={
                             colorPositionValidation === Validation.Error
-                                ? { text: 'Add color stops from 0-100 and it needs to be unique.' }
+                                ? { text: 'Add unique color stops from 0-100.' }
                                 : undefined
                         }
                         style={
--- a/packages/gradient-block/src/components/ColorFlyout.tsx
+++ b/packages/gradient-block/src/components/ColorFlyout.tsx
@@ -58,7 +58,6 @@ export const ColorFlyout = ({
 
     const setValidColorPosition = (value: string) => {
         const valueAsNumber = parseInt(value);
-        console.log(gradientColors.every((item) => item.position !== valueAsNumber));
         if (
             valueAsNumber >= 0 &&
             valueAsNumber <= 100 &&
--- a/packages/gradient-block/src/components/ColorFlyout.tsx
+++ b/packages/gradient-block/src/components/ColorFlyout.tsx
@@ -58,7 +58,12 @@ export const ColorFlyout = ({
 
     const setValidColorPosition = (value: string) => {
         const valueAsNumber = parseInt(value);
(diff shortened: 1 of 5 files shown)


packages/gradient-block/src/components/ColorFlyout.tsx
```diff
@@ -58,7 +58,11 @@ export const ColorFlyout = ({
 
     const setValidColorPosition = (value: string) => {
         const valueAsNumber = parseInt(value);
-        if (valueAsNumber >= 0 && valueAsNumber <= 100) {
+        if (
+            valueAsNumber >= 0 &&
+            valueAsNumber <= 100 &&
+            gradientColors.every((item) => item.position !== valueAsNumber)
```
celine-s, line 64: Maybe we could just add the color to the outermost part. E.g. if the valueAsNumber is 103 => add it to 100 and if it is smaller than 0 add it to 0. Atm I can still add a color to higher than 103, but it doesn't place it: https://www.loom.com/share/2d50c89f1b1f4c08bf2bb8be14a6bc04

ragi96: How can you add at position 103? I can't manage to do that

packages/gradient-block/src/components/ColorFlyout.tsx
```diff
@@ -137,7 +141,7 @@ export const ColorFlyout = ({
                     <FormControl
                         helper={
                             colorPositionValidation === Validation.Error
-                                ? { text: 'Add color stops from 0-100.' }
+                                ? { text: 'Add color stops from 0-100 and it needs to be unique.' }
```
fulopdaniel, line 144:
```suggestion
                                ? { text: 'Add unique color stops from 0-100.' }
```

fulopdaniel: Just feels a bit nicer to me but I am fine with both

fulopdaniel (approved):

packages/gradient-block/src/components/SquareBadgesRow.tsx
```diff
@@ -23,7 +23,9 @@ export const SquareBadgesRow = ({ blockWidth, gradientColors, gradientOrientatio
     };
 
     useEffect(() => {
-        prepareGradients();
+        if (gradientColors.length > 0) {
```
celine-s, line 26: maybe we could display a info text like `Please note that the gradient block requires a minimum of two colors to be selected in order to function properly.` Wdyt?

celine-s (approved):
just one idea how to solve the UX when there is no or only one color selected. Anyway now it works good - thanks!

packages/gradient-block/src/components/SquareBadgesRow.tsx
ragi96: Yeah i think that would be great, any idea where to put it?

SamuelAlev (changes requested):
Please add tests for your fixes to prevent regression

SamuelAlev (approved):

PR #539, opened 2023-04-17
feat(shared): BlockInjectButton - handle onDrop validation

celine-s:

PR #540, opened 2023-04-17
feat(shared): BlockItemWrapper allow undefined toolbar items

celine-s:


packages/shared/src/components/BlockInjectButton/types.ts
ragi96, line 1: These changes are not related to the BlockItemWrapper and are belonging to #539

celine-s: uuups, thanks! Removed

ragi96 (approved):

packages/shared/src/components/BlockItemWrapper/types.ts
```diff
@@ -5,7 +5,7 @@ import { MenuItemStyle } from '@frontify/fondue';
 export type BlockItemWrapperProps = {
     shouldHideWrapper?: boolean;
     shouldHideComponent?: boolean;
-    toolbarItems: ToolbarItem[];
+    toolbarItems: (ToolbarItem | undefined)[];
```
fulopdaniel, line 8: Shouldn't providing props in the right format would be the responsibility of the consumer? Where is this used where we pass undefined values inside the array?

celine-s: Fair enough, I see your point. It is e.g. in the Thumbnail Grid needed in the following scenario: If there is less than 2 items I don't want to show the drag and Drop handle. Therefore, I supposed also in other blocks it might be usefull to just pass undefined instead of handling the filtering in each block. We can also do it in the block. Or wdyt?

fulopdaniel (approved):

packages/shared/src/components/BlockItemWrapper/types.ts
SamuelAlev: Daniel is actually right, the responsability is on the components to not pass `undefined`

PR #541, opened 2023-04-19
feat(shared): add getBackgroundSettings

ragi96:
added `getBackgroundSettings` and used it within AssetKitBlock, and its also needed for the AnimationCurveBlock

Also extended the `getBorderRadiusSlider` so you can define a defaultRadius

### DIFF
--- a/packages/shared/src/settings/background.ts
+++ b/packages/shared/src/settings/background.ts
@@ -1,0 +1,41 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { Color, SettingBlock } from '@frontify/guideline-blocks-settings';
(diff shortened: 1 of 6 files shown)


packages/shared/src/settings/background.ts
```diff
@@ -0,0 +1,41 @@
...
+        type: 'switch',
+        defaultValue: !!options?.defaultValue,
+        on: [
+            {
+                id: colorId,
+                defaultValue: defaultColor,
+                type: 'colorInput',
+            },
+        ],
+        off: [],
```
celine-s, line 39:
```suggestion
```

celine-s (approved):
good idea. Really nice to have it!

PR #543, opened 2023-04-19
feat(shared): styles

ragi96:
Multiple Styling functions 

`getBackgroundColorStyles`
`getBorderStyles`
`getRadiusStyles`

They are already used within the AssetKitBlock and will also be used by the AnimationCurveBlock


celine-s (approved):
nice! Looking forward to reuse it in Thumbnail Grid Block as well :D

fulopdaniel:
Do you think it makes sense to add these helpers inside the utilities/react folder as they are react specific?

ragi96:
> Do you think it makes sense to add these helpers inside the utilities/react folder as they are react specific?

Yep. It definitely makes sense to move it inside there

fulopdaniel (approved):

PR #544, opened 2023-04-19
fix(DoDontItem): margin top breaks the title

fulopdaniel:
-Checkbox in settings sidebar was not visible by default
-If H3 had margin top it broke the layout


ragi96 (approved):
🚀

packages/dos-donts-block/src/DoDontItem.tsx
```diff
@@ -270,6 +270,7 @@ export const DoDontItem = React.forwardRef<HTMLDivElement, DoDontItemProps>(
                                         {
                                             ...designTokens?.heading3,
                                             marginBottom: 0,
+                                            marginTop: 0,
```
celine-s, line 273: I still have an issue with the top margin locally. If you also add it to line 263 like you did with the marginBottom it works on my side.

PR #548, opened 2023-04-24
feat(shared): add switch label to settings

celine-s:


packages/shared/src/settings/background.ts
```diff
@@ -17,17 +17,22 @@ type BackgroundSettingsType = {
...
+    hasDefaultColor?: boolean;
+    switchLabel?: string;
 };
 
 export const getBackgroundSettings = (options?: BackgroundSettingsType): SettingBlock => {
+    const showDefaultColor = options?.hasDefaultColor !== undefined ? options?.hasDefaultColor : true;
     const hasId = options?.id ? `hasBackground${options.id}` : 'hasBackground';
     const colorId = options?.id ? `backgroundColor${options.id}` : 'backgroundColor';
-    const defaultColor = options?.defaultColor || BACKGROUND_COLOR_DEFAULT_VALUE;
+    const defaultColor = showDefaultColor ? options?.defaultColor || BACKGROUND_COLOR_DEFAULT_VALUE : undefined;
```
fulopdaniel, line 28: Quick question. Why is the extra prop `hasDefaultColor` needed? Can't we just check to see if the `defaultColor` prop has been supplied?

packages/shared/src/settings/border.ts
```diff
@@ -24,6 +24,7 @@ type BorderSettingsType = {
     id?: string;
     defaultValue?: boolean;
     defaultColor?: Color;
+    switchLabel?: string;
```
ragi96, line 27: add to jsdoc

packages/shared/src/settings/background.ts
celine-s: the idea is, that you don't have to put a defaultColor. As atm we use everywhere the BACKGROUND_COLOR_DEFAULT_VALUE therefore we don't need to pass it each time. You know what I mean? I can also adapt it like you said. Wdyt?

packages/shared/src/settings/background.ts
```diff
@@ -17,17 +17,22 @@ type BackgroundSettingsType = {
     id?: string;
     defaultValue?: boolean;
     defaultColor?: Color;
+    hasDefaultColor?: boolean;
+    switchLabel?: string;
 };
```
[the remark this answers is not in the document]
celine-s: good point! added

packages/shared/src/settings/border.ts
celine-s: added 👍

packages/shared/src/settings/background.ts
```diff
@@ -10,24 +10,31 @@ import { BACKGROUND_COLOR_DEFAULT_VALUE } from './defaultValues';
  * @param options.id Custom suffix for the setting ids
  * @param options.defaultValue Default value for the background switch
  * @param options.defaultColor Default value for the background color
+ * @param options.hasDefaultColor Whether the background color should have a default value
+ * @param options.switchLabel Label for the background switch
  * @returns {SettingBlock} Returns border settings
```
ragi96, line 15:
```suggestion
 * @returns {SettingBlock} Returns background settings
```
Probably i missed that one

ragi96 (approved):

packages/shared/src/settings/background.ts
```diff
@@ -10,24 +10,31 @@ import { BACKGROUND_COLOR_DEFAULT_VALUE } from './defaultValues';
...
+    hasDefaultColor?: boolean;
+    switchLabel?: string;
 };
 
 export const getBackgroundSettings = (options?: BackgroundSettingsType): SettingBlock => {
+    const showDefaultColor = options?.hasDefaultColor !== undefined ? options?.hasDefaultColor : true;
     const hasId = options?.id ? `hasBackground${options.id}` : 'hasBackground';
     const colorId = options?.id ? `backgroundColor${options.id}` : 'backgroundColor';
-    const defaultColor = options?.defaultColor || BACKGROUND_COLOR_DEFAULT_VALUE;
+    const defaultColor = showDefaultColor ? options?.defaultColor || BACKGROUND_COLOR_DEFAULT_VALUE : undefined;
```
fulopdaniel, line 30:
```suggestion
    const defaultColor = !!options?.hasDefaultColor ? options?.defaultColor || BACKGROUND_COLOR_DEFAULT_VALUE : undefined;
```

celine-s: unfortunately not possible as hasDefaultColor is a boolean and if it is `false` we don't want to have a background color. if it is true or undefined we want to have the default color.

PR #549, opened 2023-04-24
fix(ImageBlock): fix various issues

fulopdaniel:
1. `tw-outline-line` was not applied/generated. Added back the tailwind fondue tokens to the shared tailwind config.
2. When link was added to the image and the attachments were clicked the link opened. Prevented it by only wrapping the image itself inside <a> tags.
3. Tooltips were popping up immediately making it really annoying to click the options menu in the top-right. Added some delay.
4. Download button had a huge focus outline when clicked. Looked a bit ugly. Removed it when it is clicked. Still visible by tabbing.

### DIFF
--- a/packages/image-block/src/components/Image.tsx
+++ b/packages/image-block/src/components/Image.tsx
@@ -37,7 +37,7 @@ export const ImageComponent = ({ image, blockSettings, isEditing }: ImageProps) 
                     className="tw-w-full"
                     href={link.link.link}
                     target={link.openInNewTab ? '_blank' : undefined}
-                    rel="noreferrer"
+                    rel={link.openInNewTab ? 'noopener noreferrer' : 'noreferrer'}
                 >
                     <img
                         data-test-id="image-block-img"
--- a/packages/image-block/src/components/Image.tsx
+++ b/packages/image-block/src/components/Image.tsx
@@ -27,7 +27,7 @@ type ImageProps = {
     appBridge: AppBridgeBlock;
 };
 
-export const Image = ({ image, appBridge, blockSettings, isEditing }: ImageProps) => {
+export const ImageComponent = ({ image, blockSettings, isEditing }: ImageProps) => {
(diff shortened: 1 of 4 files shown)


packages/shared/tailwind.config.js
```diff
@@ -1,6 +1,11 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
 module.exports = {
+    presets: [require('@frontify/fondue/tailwind')],
+    content: ['src/**/*.{ts,tsx}'],
```
celine-s, line 5:
```suggestion
    presets: [require('@frontify/fondue/tailwind'), require('../shared/tailwind.config')],
    content: ['src/**/*.{ts,tsx}', '../shared/src/**/*.{ts,tsx}'],
```

celine-s (approved):
nice! 
Wdyt: Could we already give it for a first review for Martin / Schirmi? Could you deploy it?

packages/shared/tailwind.config.js
fulopdaniel: This is already inside the shared folder

packages/image-block/src/components/Image.tsx
```diff
@@ -39,26 +39,34 @@ export const Image = ({ image, appBridge, blockSettings, isEditing }: ImageProps
                     target={link.openInNewTab ? '_blank' : undefined}
                     rel="noreferrer"
```
ragi96, line 40:
```suggestion
                    rel={link.openInNewTab ? 'noopener noreferrer' : 'noreferrer'}
```

Add also noopener if its target is _blank to prevent tabnapping

packages/shared/tailwind.config.js
celine-s: omg goldfish brain. :D

packages/image-block/src/components/Image.tsx
fulopdaniel: great suggestion, added

ragi96 (approved):
Great 🚀

PR #550, opened 2023-04-26
feat(Attachments): various improvements

fulopdaniel:
1. Add loading circle indicator when replacing items.
2. Supply design tokens from outside.
3. Make it accessible
4. Add keyboard support to drag n drop.

### DIFF
--- a/packages/shared/src/components/Attachments/AttachmentItem.tsx
+++ b/packages/shared/src/components/Attachments/AttachmentItem.tsx
@@ -122,7 +122,7 @@ export const AttachmentItem = forwardRef<HTMLButtonElement, AttachmentItemProps>
                 {isEditing && (
                     <div
                         data-test-id="attachments-actionbar"
-                        className="tw-flex tw-gap-[2px] tw-invisible group-focus:tw-visible group-focus-within:tw-visible group-hover:tw-visible"
+                        className="tw-flex tw-gap-0.5 tw-invisible group-focus:tw-visible group-focus-within:tw-visible group-hover:tw-visible"
                     >
                         <button
                             {...focusProps}
--- a/packages/shared/src/components/Attachments/AttachmentItem.tsx
+++ b/packages/shared/src/components/Attachments/AttachmentItem.tsx
@@ -1,6 +1,6 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
-import React, { forwardRef, useEffect, useState } from 'react';
+import React, { MutableRefObject, forwardRef, useEffect, useState } from 'react';
 import { Asset, useAssetUpload, useFileInput } from '@frontify/app-bridge';
 import { useSortable } from '@dnd-kit/sortable';
 import { useFocusRing } from '@react-aria/focus';
@@ -61,7 +61,7 @@ export const AttachmentItem = forwardRef<HTMLButtonElement, AttachmentItemProps>
         const [selectedAsset, setSelectedAsset] = useState<Asset | undefined>();
         const [openFileDialog, { selectedFiles }] = useFileInput({ multiple: true, accept: 'image/*' });
         const [uploadFile, { results: uploadResults, doneAll }] = useAssetUpload();
-        const { isFocused, focusProps } = useFocusRing();
+        const { focusProps, isFocusVisible } = useFocusRing();
 
         useEffect(() => {
             if (selectedFiles) {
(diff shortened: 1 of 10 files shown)


ragi96 (commented):
There is also one accessibility problem i noticed

https://www.loom.com/share/68eecd3c821c4fc4a30d50e35e4f89f6

packages/shared/src/components/Attachments/Attachments.spec.ct.tsx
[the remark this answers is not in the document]
fulopdaniel: Thanks ,great catch and suggestions. Fixed and added.

packages/shared/src/components/Attachments/Attachments.tsx
```diff
@@ -92,19 +94,23 @@ export const Attachments = ({
...
-                appBridge.closeAssetChooser();
+            async (result: Asset[]) => {
                 setIsFlyoutOpen(true);
+                appBridge.closeAssetChooser();
+                setAssetIdsLoading([...assetIdsLoading, toReplace.id]);
+                await onReplaceWithBrowse(toReplace, result[0]);
+                setAssetIdsLoading(assetIdsLoading.filter((id) => id !== toReplace.id));
             },
             {
                 multiSelection: false,
```
ragi96, line 105:
```suggestion
                multiSelection: false,
                selectedValueIds: internalItems.map((item) => item.id),
```
add it also on line 89, for initial adds

to prevent that one asset could be added twice, which would end up in strange behaviours at dragging, etc...

packages/shared/src/components/Attachments/AttachmentItem.tsx
```diff
@@ -107,18 +122,19 @@ export const AttachmentItem = forwardRef<HTMLDivElement, AttachmentItemProps>(
                 {isEditing && (
                     <div
                         data-test-id="attachments-actionbar"
-                        className="tw-flex tw-gap-[2px] tw-invisible group-hover:tw-visible"
+                        className="tw-flex tw-gap-[2px] tw-invisible group-focus:tw-visible group-focus-within:tw-visible group-hover:tw-visible"
```
ragi96, line 125:
```suggestion
                        className="tw-flex tw-gap-0.5 tw-invisible group-focus:tw-visible group-focus-within:tw-visible group-hover:tw-visible"
```
just minor tweak for tailwind classes

packages/shared/src/components/Attachments/Attachments.tsx
fulopdaniel: Thanks, fixed

ragi96: you missed to add it on line 107, otherwise its fine

fulopdaniel: yep, fixed

ragi96 (approved):
🚀

packages/shared/src/components/Attachments/AttachmentItem.tsx
```diff
@@ -107,18 +122,19 @@ export const AttachmentItem = forwardRef<HTMLDivElement, AttachmentItemProps>(
...
+                                'tw-bg-button-background tw-border-button-border hover:tw-bg-button-background-hover active:tw-bg-button-background-pressed tw-group tw-border tw-box-box tw-relative tw-flex tw-items-center tw-justify-center tw-outline-none tw-font-medium tw-rounded tw-h-9 tw-w-9 ',
+                                isDragging ? 'tw-cursor-grabbing' : 'tw-cursor-grab',
+                                isFocused && FOCUS_STYLE,
+                            ])}
+                        >
+                            <IconGrabHandle20 />
+                        </button>
                         <div className="-tw-mx-3">
                             <Flyout
                                 placement={FlyoutPlacement.Right}
```
SamCreasey, line 140: This flyout trigger is a nested button, which means it has two tab indexes nested inside each other. Use the render function version of the Flyout trigger to avoid this

packages/shared/src/components/Attachments/Attachments.tsx
```diff
@@ -156,12 +163,14 @@ export const Attachments = ({
                                             collisionDetection={closestCenter}
                                             onDragStart={handleDragStart}
                                             onDragEnd={handleDragEnd}
+                                            modifiers={[restrictToWindowEdges]}
```
SamCreasey, line 166: Why not restrict to Axis? There's no need for this list to be movable via x-axis

SamCreasey:
Screenshot 2023-05-04 at 13 43 07
Double Focus rings are visible

SamCreasey:
https://www.loom.com/share/c5ac3ddb1d344468906bd20d34fb3bba. Some kind of bug with the drag state if you click to close the flyout before the drag animation finishes

SamCreasey (changes requested):

packages/shared/src/components/Attachments/Attachments.tsx
fulopdaniel: I fiddled around with this and if I fix it to y axis it results in a worse UX, at least that was my experience. I like that I can slightly move the drag overlay outside its axis, feels a bit more natural.

What do you think?

packages/shared/src/components/Attachments/AttachmentItem.tsx
fulopdaniel: Thanks, fixed!

PR #552, opened 2023-05-03
feat(shared): add security settings

ragi96:
[Task](https://app.clickup.com/t/8677unvhw) 

Security Settings if asset is download

added on
- audio-block
- image-block

### DIFF
--- a/packages/audio-block/src/AudioBlock.spec.ct.tsx
+++ b/packages/audio-block/src/AudioBlock.spec.ct.tsx
@@ -136,7 +136,7 @@ describe('Audio Block', () => {
         cy.get(AudioBlockSelector).should('have.css', 'flex-direction', 'column');
     });
 
-    it('renders an audio block without download button as its set in the security settings', () => {
+    it('renders an audio block with download button as its set in the security settings', () => {
         const asset = AssetDummy.with(312);
         const [AudioBlockWithStubs] = withAppBridgeBlockStubs(AudioBlock, {
             blockSettings: {
--- a/packages/audio-block/src/AudioBlock.spec.ct.tsx
+++ b/packages/audio-block/src/AudioBlock.spec.ct.tsx
@@ -150,4 +150,19 @@ describe('Audio Block', () => {
         mount(<AudioBlockWithStubs />);
         cy.get(DownloadButtonSelector).should('exist');
     });
+
+    it('renders an audio block without download button as its not allowed by the security settings', () => {
+        const asset = AssetDummy.with(312);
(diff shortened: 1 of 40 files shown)


packages/audio-block/src/settings.ts
```diff
@@ -3,11 +3,14 @@
 import { FileExtensionSets } from '@frontify/app-bridge';
 import { AssetChooserObjectType, IconEnum, defineSettings } from '@frontify/guideline-blocks-settings';
 
-import { AudioSecurity, TextPosition } from './types';
+import { TextPosition } from './types';
+import {
+    getSecurityDownloadableSetting,
+    getSecurityGlobalControlSetting,
+} from '@frontify/guideline-blocks-shared/src/settings/security';
```
fulopdaniel, line 10:
```suggestion
} from '@frontify/guideline-blocks-shared'
```

packages/audio-block/src/AudioBlock.spec.ct.tsx
```diff
@@ -133,4 +135,19 @@ describe('Audio Block', () => {
         mount(<AudioBlockWithStubs />);
         cy.get(AudioBlockSelector).should('have.css', 'flex-direction', 'column');
     });
+
+    it('renders an audio block without download button as its set in the security settings', () => {
```
fulopdaniel, line 139:
```suggestion
    it('renders an audio block with download button as its set in the security settings', () => {
```

fulopdaniel: Can you also add a test for not rendering it?

packages/image-block/src/ImageBlock.spec.ct.tsx
```diff
@@ -50,8 +51,12 @@ describe('Image Block', () => {
         cy.get(ImageBlockImageSelector).should('exist');
     });
 
-    it('should render the download button if the image is uploaded', () => {
+    it('should render the download button if the image is uploaded and security settings allows it', () => {
```
fulopdaniel, line 54: Can you add an extra test case checking if the security does not allow it?

packages/audio-block/src/AudioBlock.spec.ct.tsx
ragi96: yep, added

packages/image-block/src/ImageBlock.spec.ct.tsx
ragi96: also added one more test here

fulopdaniel (approved):

fulopdaniel:
🥇

Walkeezy (approved):

PR #554, opened 2023-05-03
feat: added i18nFields and searchFields

ragi96:
[Task](https://app.clickup.com/t/8677vgt82)
checked all blocks and added missing fields to the `manifest.json`


triggertoo (approved):
🚀

PR #556, opened 2023-05-04
feat(AnimationCurveBlock): ux review 

ragi96:
[Task](https://app.clickup.com/t/865c7g96v)


packages/animation-curve-block/src/components/AnimationCurveFlyout.tsx
```diff
@@ -137,75 +134,76 @@ export const AnimationCurveFlyout = ({
...
                                 value={localAnimationFunction.parameters.y2.toString()}
                                 required
+                                decorator="Y"
                             />
                         </div>
                     </div>
                     <div className="tw-w-20 tw-relative">
                         <TextInput
+                            onFocus={() => setZIndex(0)}
+                            onBlur={() => setZIndex(1000)}
```
fulopdaniel, line 206: I'm sure I'm missing something but if the user starts to edit the input value, shouldn't it have higher zIndex than previously to be on top?

packages/animation-curve-block/src/AnimationCurveBlock.spec.ct.tsx
```diff
@@ -474,4 +497,16 @@ describe('AnimationCurve Block', () => {
         cy.get(TEXT_INPUT_SELECTOR).eq(2).clear().realPress('ArrowDown');
         cy.get(FLYOUT_SELECTOR).find(BUTTON_SELECTOR).last().should('be.disabled');
     });
+
+    it('should show the correct cubic-bezier values of an Eas In curve ', () => {
```
Walkeezy, line 501: Should this be like this?
```suggestion
    it('should show the correct cubic-bezier values of an easing curve ', () => {
```
Or what does it mean?

packages/shared/src/components/BlockItemWrapper/BlockItemWrapper.spec.ct.tsx
```diff
@@ -115,4 +115,22 @@ describe('Block Item Wrapper', () => {
         cy.get(BlockItemWrapperSelector).focus();
         cy.get(ToolbarSelector).should('be.visible');
     });
+
+    it('should render the outline and the toolbar if should be shown is true', () => {
```
Walkeezy, line 119: Nit: ```suggestion
    it('should render the outline and the toolbar if enabled', () => {
```

Walkeezy (approved):

packages/animation-curve-block/src/AnimationCurveBlock.spec.ct.tsx
ragi96: Well its real cubic-bezier values of an `Eas In Curve`.

packages/animation-curve-block/src/components/AnimationCurveFlyout.tsx
ragi96: Well the zIndex which get set is for the container around the animation curve and not for the TextInput, so onFocus the  the animation curve (basically just the handles can overflow) gets behind the TextInputs. But as its not clear I will adjust the naming a bit so its hopefully clearer

fulopdaniel (approved):

PR #557, opened 2023-05-08
feat: add Glyphs block

Walkeezy:
Ticket: https://app.clickup.com/t/863ghy2bx


packages/glyphs-block/src/GlyphsBlock.tsx
```diff
@@ -0,0 +1,73 @@
...
+                fontSize: blockSettings.fontSize,
+                fontFamily: fontFamily || 'inherit',
+                color: toRgbaString(fontColor),
+            }}
+        >
+            <ul
+                className="tw-grid tw-grid-cols-6"
+                style={{
+                    gap: borderWidth,
+                    padding: borderWidth,
```
celine-s, line 66: I think you could remove the outer div and move all in `ul`

packages/glyphs-block/src/GlyphsBlock.tsx
```diff
@@ -0,0 +1,73 @@
...
+                    ...(hasBorder && {
+                        outlineStyle: borderStyle,
+                        outlineWidth: borderWidth,
+                        outlineColor: toRgbaString(borderColor),
+                    }),
+                    ...getRadiusStyles(radiusChoice, hasRadius, radiusValue),
+                }}
+            >
+                {char}
+                {isLetter && char.toLowerCase()}
```
celine-s, line 47: should it also do it with a lowercase letter? or maybe add a check in `isLetter` if it's uppercase. Wdyt?

packages/glyphs-block/src/helpers.ts
```diff
@@ -0,0 +1,37 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { Choice } from '@frontify/guideline-blocks-settings';
+import { ApiFontStyle } from './types';
+
+export const getFonts = async (projectId: number): Promise<Choice[]> => {
+    const response = await fetch(`/api/project-font-family?project_id=${projectId}`);
+    const responseJson = await response.json();
+    const fonts = responseJson.data.map((fontStyle: ApiFontStyle) => ({
```
celine-s, line 9: add typing here to avoid any.

like:

```
const fonts: { label: string; value: string }[] = responseJson.data.map((fontStyle: ApiFontStyle) => ({
        label: fontStyle.name,
        value: fontStyle.css_family_name,
    }));

    return fonts.filter((font, index, self) => self.findIndex(({ value }) => value === font.value) === index);
    ```

File: packages/glyphs-block/src/helpers.ts [hunk 4]
```diff
@@ -0,0 +1,37 @@
...
+        return {
+            label: fontStyle.font_weight,
+            value: fontStyle.font_weight,
+        };
+    });
+    const uniqueFontWeights = fontWeights.filter(
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        (font: { value: any }, index: any, self: any[]) => self.findIndex((f) => f.value === font.value) === index
+    );
+    return uniqueFontWeights;
```
celine-s, line 36:
```suggestion
const fontStyles: ApiFontStyle[] = responseJson.data.filter(
        (fontStyle: ApiFontStyle) => fontStyle.name === font && fontStyle.font_style === 'normal'
    );
    const fontWeights: { label: string; value: string }[] = fontStyles.map(({ font_weight }) => ({
        label: font_weight,
        value: font_weight,
    }));

    return fontWeights.filter((font, index, self) => self.findIndex(({ value }) => value === font.value) === index);
```

packages/glyphs-block/src/settings.ts
```diff
@@ -0,0 +1,89 @@
...
+        {
+            id: 'font',
+            type: 'multiInput',
+            label: 'Font selection',
+            lastItemFullWidth: true,
+            blocks: [
+                {
+                    id: 'fontWeight',
+                    type: 'dropdown',
+                    defaultValue: 'normal',
```
celine-s, line 47: I don't see anything by changing from 400 to normal with my font. Probably just my font. still wanted to check. Does it work with one of yours?

packages/glyphs-block/src/GlyphsBlock.tsx
Walkeezy: Current regex check also checks if it's an uppercase letter and only does it to uppercase letters. This is the current wanted behaviour as far as I know.

packages/glyphs-block/src/settings.ts
Walkeezy: This is because `400 = normal`. It depends on the font which font weights are available.

packages/glyphs-block/src/GlyphsBlock.tsx
Walkeezy: Changed 👍🏻

packages/glyphs-block/README.md
```diff
@@ -0,0 +1 @@
+# Frontify Glyphs Block
```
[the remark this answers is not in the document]
Walkeezy: Added

packages/glyphs-block/src/helpers.ts
Walkeezy: Optimized it 👍🏻

fulopdaniel (approved):

fulopdaniel:
Looks good to meeee 🚀 🚀

packages/glyphs-block/src/settings.ts
celine-s: makes sense 👍

celine-s (approved):
nice! 🚀

packages/glyphs-block/src/GlyphsBlock.tsx
celine-s: ah true - perf 👍

PR #559, opened 2023-05-09
feat(animationCurveBlock): second round of ux fixes 

ragi96:
[Task](https://app.clickup.com/t/865c7g96v)

Shared:
- added getGutterSettings (used within Text-Block and Animation-Curve-Block)

### DIFF
--- a/packages/animation-curve-block/src/AnimationCurveBlock.spec.ct.tsx
+++ b/packages/animation-curve-block/src/AnimationCurveBlock.spec.ct.tsx
@@ -3,11 +3,11 @@
 import { mount } from 'cypress/react';
 
 import { withAppBridgeBlockStubs } from '@frontify/app-bridge';
-import { BorderStyle, Radius, convertToRteValue } from '@frontify/guideline-blocks-shared';
+import { BorderStyle, GutterSpacing, Radius, convertToRteValue } from '@frontify/guideline-blocks-shared';
 import { TextStyles } from '@frontify/fondue';
 
 import { AnimationCurveBlock } from './AnimationCurveBlock';
-import { AnimationCurve, AnimationCurveType, Spacing } from './types';
+import { AnimationCurve, AnimationCurveType } from './types';
 
 const BLOCK_SELECTOR = '[data-test-id="animation-curve-block"]';
 const CARD_SELECTOR = '[data-test-id="animation-curve-card"]';
@@ -79,7 +79,7 @@ describe('AnimationCurve Block', () => {
         const [AssetKitBlockWithStubs] = withAppBridgeBlockStubs(AnimationCurveBlock, {
             blockSettings: {
                 columns: 2,
-                spacingChoice: Spacing.Large,
+                spacingChoice: GutterSpacing.Large,
             },
         });
         mount(<AssetKitBlockWithStubs />);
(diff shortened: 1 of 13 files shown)


packages/shared/src/settings/gutter.ts
```diff
@@ -0,0 +1,72 @@
...
+import { SettingBlock, appendUnit, numericalOrPixelRule, presetCustomValue } from '@frontify/guideline-blocks-settings';
+import { GutterSpacing, gutterSpacingStyleMap } from './types';
+
+/**
+ * Returns background settings: background switch, background color
+ *
+ * @param options Options for the settings
+ * @param options.id Custom id for the gutters switch
+ * @returns {SettingBlock} Returns background settings
+ */
```
Walkeezy, line 12: Is this correct? Background settings?

ragi96: No copy pasta missed to adjust it

packages/animation-curve-block/src/AnimationCurveBlock.spec.ct.tsx
```diff
@@ -152,20 +152,6 @@ describe('AnimationCurve Block', () => {
...
-                hasParameter: true,
-            },
-        });
-        mount(<AssetKitBlockWithStubs />);
-
-        cy.get(CARD_PARAMETERS_SELECTOR).should('css', 'cursor', 'copy');
-        cy.get(CARD_PARAMETERS_SELECTOR).click();
-        cy.get(CARD_PARAMETERS_SELECTOR).should('css', 'cursor', 'auto');
-    });
-
```
Walkeezy, line 168: Was this test removed intentionally?

Walkeezy (approved):
Works well!

packages/animation-curve-block/src/components/AnimationCanvas.tsx
```diff
@@ -105,18 +105,27 @@ export const AnimationCanvas = ({
...
-            }
+            setAnimationFunction && setAnimationFunction(updatedAnimationFunction(animationValues, draggingPoint));
         },
         // eslint-disable-next-line react-hooks/exhaustive-deps
         [draggingPoint, viewBox.width, viewBox.height]
     );
 
+    const calculateXValue = (viewBoxX: number, viewBoxWidth: number) => {
+        const x = viewBoxX / viewBoxWidth;
+        if (x < 0) {
```
fulopdaniel, line 120: Optional, to consume less lines:
`return Math.max(Math.min(x, 1), 0)`

packages/animation-curve-block/src/components/AnimationCurveFlyout.tsx
```diff
@@ -93,10 +93,9 @@ export const AnimationCurveFlyout = ({
 
     return (
         <Flyout
-            trigger={<></>}
+            trigger={(props, ref) => <div ref={ref as MutableRefObject<HTMLDivElement>} />}
```
fulopdaniel, line 96: What is the purpose of this change?

ragi96: Its to get rid of the trigger button which would be focusable but as its open via the toolbar this would be quite confusing for the user.

Or described by Martin:
`Accessibility: Edit animation button seems to be targeted twice when going through with tab. Second time there is no visual highlight, which could be confusing.`

packages/animation-curve-block/src/components/AnimationCanvas.tsx
ragi96: Thanks. Adjusted and got rid of that function

fulopdaniel (approved):

PR #560, opened 2023-05-11
feat(RTE): allow document paths

fulopdaniel:
1. Prepend URL with https:// if not exists.
2. Allow adding URLs that begin with /document/....

### DIFF
--- a/packages/shared/src/components/RichTextEditor/plugins/LinkPlugin/index.ts
+++ b/packages/shared/src/components/RichTextEditor/plugins/LinkPlugin/index.ts
@@ -7,8 +7,11 @@ import { CustomFloatingLink } from './FloatingLink/CustomFloatingLink';
 import { LINK_PLUGIN } from './id';
 import { LinkButton } from './LinkButton';
 import { LinkMarkupElement } from './LinkMarkupElement';
+import { relativeUrlRegex, telOrMailRegex, urlRegex } from './utils';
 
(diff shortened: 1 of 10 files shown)


Walkeezy (approved):
Nice!

packages/shared/src/components/RichTextEditor/plugins/LinkPlugin/index.ts
```diff
@@ -7,8 +7,11 @@ import { CustomFloatingLink } from './FloatingLink/CustomFloatingLink';
 import { LINK_PLUGIN } from './id';
 import { LinkButton } from './LinkButton';
 import { LinkMarkupElement } from './LinkMarkupElement';
+import { relativeUrlRegex, telOrMailRegex, urlRegexWithHttps } from './utils';
 
-export const isValidUrl = (url: string): boolean => !!url.match(/^((https?:\/\/|mailto:|tel:).+)|\/r\/.+/);
+export const isValidUrl = (url: string): boolean => {
```
ragi96, line 12: Also that function is worth to have some unit tests

fulopdaniel: Thanks, done!

packages/shared/src/components/RichTextEditor/plugins/LinkPlugin/utils/regex.ts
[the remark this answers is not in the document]
fulopdaniel: Good idea, I even caught a bug when implementing tests. Done!

PR #562, opened 2023-05-16
feat(GlyphsBlock): adjust border, radius & settings

Walkeezy:
Fixes for design review: https://app.clickup.com/t/863ghy2bx

### DIFF
--- a/packages/glyphs-block/src/GlyphsBlock.tsx
+++ b/packages/glyphs-block/src/GlyphsBlock.tsx
@@ -1,8 +1,9 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
 import { useBlockSettings } from '@frontify/app-bridge';
-import { getRadiusStyles, toRgbaString } from '@frontify/guideline-blocks-shared';
+import { toRgbaString } from '@frontify/guideline-blocks-shared';
 import { ReactElement } from 'react';
+import { getRadiusValue } from './helpers';
 import { DEFAULT_CHARS } from './settings';
 import { BlockProps, Settings } from './types';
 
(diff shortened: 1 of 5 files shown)


packages/glyphs-block/src/GlyphsBlock.tsx
```diff
@@ -40,7 +46,18 @@ export const GlyphsBlock = ({ appBridge }: BlockProps): ReactElement => {
...
+                    }),
+                    ...(isSixthChar && {
+                        borderTopRightRadius: getRadiusValue(radiusChoice, hasRadius, radiusValue),
+                    }),
+                    ...(isBottomLeftChar && {
+                        borderBottomLeftRadius: getRadiusValue(radiusChoice, hasRadius, radiusValue),
+                    }),
+                    ...(isBottomRightChar && {
+                        borderBottomRightRadius: getRadiusValue(radiusChoice, hasRadius, radiusValue),
+                    }),
```
celine-s, line 60: this now gets a bit big :D. you could use `Object.assign()`. Wdyt?

Like: 
const style = Object.assign(
            {},
            hasBackground && { backgroundColor: toRgbaString(backgroundColor) },
...
}

packages/glyphs-block/src/GlyphsBlock.tsx
```diff
@@ -40,7 +46,18 @@ export const GlyphsBlock = ({ appBridge }: BlockProps): ReactElement => {
                         outlineWidth: borderWidth,
```
celine-s, line 46: you could take all the border infos into one css: `outline: `${borderWidth} ${borderStyle} ${toRgbaString(borderColor)}`,`

Walkeezy: Changed 👍🏻

packages/glyphs-block/src/GlyphsBlock.tsx
Walkeezy: I think it's easier understandable without Object.assign(), but moved the styles to a const to make it a bit cleaner.

celine-s (approved):
🚀

ragi96 (approved):
🚀

PR #565, opened 2023-05-23
feat(GlyphBlock): remove upper/lowercase logic, fix border radius logic

Walkeezy:
- Remove logic to automatically add lowercase for uppercase letters
- Fix border radius logic
- Change default value

### DIFF
--- a/packages/glyphs-block/src/GlyphsBlock.tsx
+++ b/packages/glyphs-block/src/GlyphsBlock.tsx
@@ -25,13 +25,12 @@ export const GlyphsBlock = ({ appBridge }: BlockProps): ReactElement => {
         radiusValue,
     } = blockSettings;
 
-    const splittedItems = chars.split(',');
+    const splittedItems = chars.split(',').map((char) => char.trim());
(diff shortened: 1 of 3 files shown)


packages/glyphs-block/src/GlyphsBlock.tsx
```diff
@@ -25,13 +25,12 @@ export const GlyphsBlock = ({ appBridge }: BlockProps): ReactElement => {
         radiusValue,
     } = blockSettings;
 
-    const splittedItems = chars.split(',');
+    const splittedItems = chars.split(',').map((char) => char.trim());
+    const numberOfRows = Math.ceil(splittedItems.length / 6);
```
ragi96, line 29: Can we put that `6` into constant like `ITEMS_PER_ROW`

celine-s (approved):
nice!

packages/glyphs-block/src/GlyphsBlock.tsx
Walkeezy: Good point, added 👍🏻

ragi96 (approved):

PR #566, opened 2023-05-23
feat(RTE): remove emoji plugin

celine-s:
bug ticket: https://app.clickup.com/t/2523021/TASK-7268

My suggestion is to just remove the plugin for the moment and if we want to add it make a ticket which should improve the performance of the plugin. The behaviour is also the same on [plate](https://plate.udecode.io/docs/playground). 
Wdyt @ragi96 ?

### DIFF
--- a/packages/text-block/src/getPlugins.tsx
+++ b/packages/text-block/src/getPlugins.tsx
@@ -10,7 +10,6 @@ import {
     BreakAfterPlugin,
     CheckboxListPlugin,
     CodePlugin,
-    EmojiPlugin,
     ItalicPlugin,
     OrderedListPlugin,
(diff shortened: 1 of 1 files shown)


Walkeezy:
Is the issue only in the text block or is the emoji plugin also in other blocks?

ragi96:
Removed it here https://github.com/Frontify/guideline-blocks/pull/561 

It was the only appearance left,  so all blocks would be without the emoji plugin

celine-s:
> Removed it here #561
> 
> It was the only appearance left, so all blocks would be without the emoji plugin

Thats why I made the PR to this branch ;). Thanks for takling it 👍

PR #567, opened 2023-05-23
fix(calloutBlock): show custom icon when selected

celine-s:


Walkeezy (approved):

ragi96 (approved):
Please keep in mind to not do a community release because of the theme settings.

PR #569, opened 2023-05-25
fix(callout): overriding the theme settings variables

ragi96:
before the override was on `:root` and so it has overridden all styles on the page, and if you would have multiple on a page the last one would be taken - now the overrides are just for the specific block.

[Task](https://app.clickup.com/t/867814fnp)

[Check on product](https://product.frontify.com/document/11#/new-content-blocks/callout-block)

### DIFF
--- a/packages/callout-block/src/CalloutBlock.tsx
+++ b/packages/callout-block/src/CalloutBlock.tsx
@@ -75,8 +75,6 @@ export const CalloutBlock: FC<BlockProps> = ({ appBridge }) => {
 
     const onTextChange = (value: string) => value !== blockSettings.textValue && setBlockSettings({ textValue: value });
 
-    //document.documentElement.style.setProperty(`${THEME_PREFIX}heading1-color`, textColor);
-
     return (
         <>
             <style type="text/css">{`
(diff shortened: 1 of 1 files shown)


SamuelAlev (changes requested):
Request changes just to get what does "it has overridden all styles on the page" mean, what style did it override?

fulopdaniel:
> Request changes just to know what does "it has overridden all styles on the page" mean, what style did it override?

I suppose the heading styles, as the callout block can set custom heading colors.

fulopdaniel (approved):

ragi96:
> Request changes just to know what does "it has overridden all styles on the page" mean, what style did it override?

Yes this block adjusts the the font color of the the textStyles based on the background color, so there is always a sufficient contrast

celine-s (commented):
Initially by adding a new block the text color is wrong: https://www.loom.com/share/6f58365207c04d44987b546648f97f47.

Changing sth in settings => correct color

ragi96:
> Initially by adding a new block the text color is wrong: https://www.loom.com/share/6f58365207c04d44987b546648f97f47.
> 
> Changing sth in settings => correct color

added a workaround and created a [task](https://app.clickup.com/t/86781grd1) to investigate it deeper

celine-s (approved):
lgtm

PR #578, opened 2023-06-01
fix(RTE): link validation for mailto/tel links and longer paths

celine-s:
[ticket](https://app.clickup.com/t/86781gu71)

* https:// was added to mailto & tel (e.g. mailto:info@frontify.com => got https://mailto:info@frontify.com)
* `https://github.com/Frontify/guideline-blocks/pull/578` was not considered as a correct link


Walkeezy (approved):
🥇

packages/shared/src/components/RichTextEditor/plugins/LinkPlugin/utils/regex.ts
```diff
@@ -1,6 +1,6 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
-export const telOrMailRegex = /^(mailto:|tel:)[\w%+.-@]+$/;
+export const telOrMailRegex = /^(mailto:|tel:)([\w %+.@\-]+)$/i;
 export const urlRegex = /[^\s#$./?].\S*\.[a-z]{2,}$/i;
 export const urlRegexWithHttps = /^(https?:\/\/)[^\s#$./?].\S*\.[a-z]{2,}$/i;
```
fulopdaniel, line 5: Is this still used somewhere? If not it can be removed.

celine-s: good point! I removed them and changed the addHttps such that it uses the same logic as the validation

fulopdaniel (approved):

PR #579, opened 2023-06-02
fix(CalloutBlock): prevent text color overwrite

Walkeezy:
Paragraph text in Callout block get's overwritten by hub appearance css, this should fix it.

I also refactored the block a bit: removed the `style` tag for the css variables and directly added them to the element itself.

Last but not least: added a bunch of tests for all of the color handling.

Before:
Bildschirmfoto 2023-06-02 um 08 41 55

After:
Bildschirmfoto 2023-06-02 um 08 42 11


fulopdaniel (approved):

fulopdaniel:
very nice

celine-s (approved):
nice, thanks!

PR #580, opened 2023-06-05
fix(RteSerializer): Button styles

Walkeezy:
Fixes an issue where a callout block would not appear in view mode, if there is a button inside the content. Ticket: https://app.clickup.com/t/2523021/TASK-7617

This was caused because the serializer didn't get the correct button styles passed down. And also because of a call to the hover property, which was undefined: 
Bildschirmfoto 2023-06-05 um 09 26 12


fulopdaniel (approved):

celine-s (approved):
🚀

PR #584, opened 2023-06-06
feat(CalloutBlock): adjust for light colors

Walkeezy:
When a light color (white, lightblue etc.) is defined in the theme settings, the text color would be unreadable in the callout block with `appearance: light` (see issue mentioned [here](https://app.clickup.com/t/2523021/TASK-7595?comment=90100014900142&threadedComment=90100014912760) and/or loom [here](https://www.loom.com/share/ab1c45bb3f3b44d4822966a24ca57c78)).

This PR adds a new utility, which returns a darkened text color with a contrast ratio of at least 4.5 to the background. Also added some test for it.


celine-s (approved):
Lgtm
Did you check the new text color with the designer?

Walkeezy:
> Lgtm Did you check the new text color with the designer?

Yes, discussed it with Martin on how we should go about this and will also do a UX review before releasing the block.

fulopdaniel (approved):
nice!

PR #585, opened 2023-06-06
feat(RTE): save on change

fulopdaniel:


packages/shared/src/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -15,19 +15,36 @@ export const RichTextEditor = ({
...
+
     if (isEditing) {
-        const saveText = (newValue: string) => newValue !== value && onBlur(newValue);
         return (
             <FondueRichTextEditor
                 id={id}
                 value={value}
                 border={false}
                 placeholder={placeholder}
-                onBlur={saveText}
```
SamCreasey, line 29: Is removing this function an issue? I imagine it was implemented for a specific reason

SamCreasey (approved):
Works like a charm 🥳

packages/text-block/src/TextBlock.tsx
```diff
@@ -15,7 +15,17 @@ export const TextBlock = ({ appBridge }: BlockProps): ReactElement => {
     const [blockSettings, setBlockSettings] = useBlockSettings<Settings>(appBridge);
     const { content, columnNumber, columnGutterSimple, columnGutterCustom, isColumnGutterCustom } = blockSettings;
     const gap = isColumnGutterCustom ? columnGutterCustom : spacingValues[columnGutterSimple];
-    const onTextChange = (content: string) => setBlockSettings({ content });
+    const [isApiRequestPending, setIsApiRequestPending] = useState(false);
+
+    const onTextChange = (newContent: string) => {
```
celine-s, line 20: Could you also add this to the callout block?

celine-s (commented):
Awesome!

packages/shared/src/components/RichTextEditor/RichTextEditor.spec.ct.tsx
Walkeezy, line 1: Very nice work! Could you add a test case for `shouldPreventPageLeave` either directly here or on the blocks (Text & callout) itself?

fulopdaniel: I am not sure how to test page reload in component tests, maybe end to end tests are better suitable for checking what happens if the page refresh button is pressed. What do you think, do you have any suggestions? In cypress I can do cy.reload() but then no dialog pops up, it just hangs. Also not sure how this would behave in the CI

Walkeezy: Let's add it when we have e2e tests 👍🏻

Walkeezy (approved):

PR #588, opened 2023-06-07
fix(CalloutBlock): text alignment

Walkeezy:
- Fix horizontal alignment of text, which wasn't working in view mode, because the content wasn't taking over 100% of the width
- Request from Martin: Remove margin of last text element inside the block, so that the content is also vertically centered inside the block

Before:
Bildschirmfoto 2023-06-07 um 13 05 25

After:
Bildschirmfoto 2023-06-07 um 13 05 45

### DIFF
--- a/packages/callout-block/src/CalloutBlock.tsx
+++ b/packages/callout-block/src/CalloutBlock.tsx
@@ -52,7 +52,11 @@ export const CalloutBlock = ({ appBridge }: BlockProps): ReactElement => {
     const textColor = getTextColor(blockSettings.appearance, accentColor, backgroundColor);
 
     const textDivClassNames = joinClassNames([
-        'callout-block tw-flex tw-items-center',
+        'tw-flex tw-items-center',
+        '[&>div>*:last-child]:tw-my-0', // Remove margin from last child in view mode
+        '[&>div>*:last-child>span]:!tw-my-0',
+        '[&>div>div>*:last-child]:tw-my-0', // Remove margin from last child in edit mode
+        '[&>div>div>*:last-child>span]:!tw-my-0',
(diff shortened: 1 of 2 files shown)


packages/callout-block/src/CalloutBlock.tsx
```diff
@@ -117,6 +117,19 @@ export const CalloutBlock = ({ appBridge }: BlockProps): ReactElement => {
                     plugins={getDefaultPluginsWithLinkChooser(appBridge)}
                 />
             </div>
+
+            {/* Hacky way to remove margin from the last element, so that the text is vertically centered in the callout block */}
```
celine-s, line 121: Shouldn't it also remove margin-top from the first-child?

Walkeezy: Good point, added 👍🏻

packages/callout-block/src/CalloutBlock.tsx
```diff
@@ -117,6 +117,19 @@ export const CalloutBlock = ({ appBridge }: BlockProps): ReactElement => {
...
+            <style>{`
+                .callout-block > div > *:last-child,
+                .callout-block > div > div > *:last-child {
```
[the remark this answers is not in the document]
Walkeezy: Aah yeah, changed it to tailwind. Didn't think it was possible ... 😄

fulopdaniel (approved):

packages/callout-block/src/CalloutBlock.tsx
celine-s: I don't see any classnames on first-child. Did you forget to commit it? Or I'm I missing sth?
image

celine-s: 👍

celine-s (approved):

PR #589, opened 2023-06-09
feat(AudioBlock): add `shouldPreventPageLeave` to RTE & refactor

Walkeezy:
Ticket: https://app.clickup.com/t/867838pj1


packages/audio-block/src/AudioBlock.tsx
```diff
@@ -120,6 +83,26 @@ export const AudioBlock = ({ appBridge }: BlockProps) => {
...
+        } else {
+            setBlockSettings({ title: value }).finally(() => {
+                setIsApiRequestPending(false);
+            });
+        }
+    };
+
+    const onDescriptionChange = (value: string) => {
+        if (value === blockSettings.description) {
+            setIsApiRequestPending(false);
```
fulopdaniel, line 98: small edgecase that might be worth considering: Title is edited, then user clicks in the description and then refreshes the page. This case the data might be lost because the onDescriptionChange function is called on click as well. That would set the isApiRequestPending to false even if the previous request is still pending.

Solution: Use separate variables for all settings fields. Or an object.

What do you think?

celine-s (commented):
I somehow have issue with saving the title (see [loom](https://www.loom.com/share/7dee36f3d0164ea0a9f98472a7051103)). Does that work on your end?

Walkeezy:
> I somehow have issue with saving the title (see [loom](https://www.loom.com/share/7dee36f3d0164ea0a9f98472a7051103)). Does that work on your end?

No, can't reproduce the issue 🤔

celine-s (approved):
Nice work!

ragi96 (approved):

PR #590, opened 2023-06-12
feat(CalloutColors): use custom brightness

Walkeezy:
Add option to add custom threshold to `isDark` as discussed with Martin

### DIFF
--- a/packages/callout-block/src/helpers/getTextColor.ts
+++ b/packages/callout-block/src/helpers/getTextColor.ts
@@ -1,10 +1,14 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
-import { getReadableColor, isDark } from '@frontify/guideline-blocks-shared';
+import { getReadableColor, toShortRgba } from '@frontify/guideline-blocks-shared';
+import tinycolor from 'tinycolor2';
 import { Appearance } from '../types';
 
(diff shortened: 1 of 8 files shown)


packages/callout-block/src/helpers/getTextColor.ts
```diff
@@ -1,10 +1,14 @@
...
-import { getReadableColor, isDark } from '@frontify/guideline-blocks-shared';
+import { getReadableColor, toShortRgba } from '@frontify/guideline-blocks-shared';
+import tinycolor from 'tinycolor2';
 import { Appearance } from '../types';
 
 export const getTextColor = (appearance: Appearance, color: string, backgroundColor: string): string => {
-    const isDarkColor = isDark(color);
+    const parsedColor = tinycolor(toShortRgba(color));
+    const brightness = parsedColor.getBrightness();
+    const isDarkColor = brightness < 150;
```
fulopdaniel, line 10: Can this logic be in the 'shared' folder, so we don't need to add tinycolor to the callout dependencies?

Walkeezy: Refactored it to have a threshold option on the `isDark` utility 👍🏻

celine-s (approved):
nice!

fulopdaniel (approved):

PR #591, opened 2023-06-12
feat(ImageBlock): fix stylings, refactoring and add Rte save on change

celine-s:


packages/image-block/src/ImageBlock.tsx
```diff
@@ -14,7 +14,7 @@ import {
 import '@frontify/fondue-tokens/styles';
 import { BlockProps } from '@frontify/guideline-blocks-settings';
 import 'tailwindcss/tailwind.css';
-import { Settings, mapCaptionPositionClasses } from './types';
```
Walkeezy, line 17: Shouldn't they be named `ratioValues`?

packages/image-block/src/components/Image.tsx
```diff
@@ -28,8 +20,20 @@ type ImageProps = {
...
+export const ImageComponent = ({ image, blockSettings, isEditing, appBridge }: ImageProps) => {
+    const link = blockSettings?.hasLink && blockSettings?.linkObject?.link && blockSettings?.linkObject;
+    const imageStyle = getImageStyle(blockSettings, image.width);
+
+    const Image = (
+        <img
+            data-test-id="image-block-img"
+            className="tw-flex tw-w-full"
+            loading="lazy"
+            src={image.genericUrl.replace('{width}', `${800 * window.devicePixelRatio}`)}
```
Walkeezy, line 32: Is it possible that `window.devicePixelRatio` is undefined or null on some browsers/devices?

packages/image-block/src/components/ImageCaption.tsx
Walkeezy, line 1: Selected image title and caption always show as body text in the RichTextEditor selection.

Bildschirmfoto 2023-06-20 um 07 41 59

celine-s: good catch! fixed

packages/image-block/src/components/Image.tsx
celine-s: yes indeed that could be - thanks! Added a condition also in the other blocks where it was used.

Walkeezy (approved):

packages/image-block/src/settings.ts
ragi96, line 1: As discussed with Martin remove `Image` from `Image Padding`

celine-s: done 👍

ragi96 (approved):

PR #595, opened 2023-06-16
feat(shared): DnD accessibility

ragi96:


fulopdaniel (approved):
Nice work

packages/animation-curve-block/src/components/Card.tsx
```diff
@@ -87,7 +88,13 @@ export const Card = forwardRef<HTMLDivElement, CardProps>(
                     shouldHideWrapper={replaceWithPlaceholder || !isEditing}
                     shouldHideComponent={replaceWithPlaceholder}
                     toolbarItems={[
-                        { icon: <IconArrowMove16 />, tooltip: 'Drag to move', draggableProps },
+                        {
+                            icon: <IconArrowMove16 />,
+                            tooltip: 'Drag or press ↵ to move',
+                            draggingTooltip: 'Move with ↑↓←→ and confirm with ↵',
```
celine-s, line 94: are those tooltips from the design? I'm not sure if it makes sense to keep them like this, as those actions should be imho self-explanatory. We don't do it e.g. on a delete button "press enter to delete". Wdyt?

packages/thumbnail-grid-block/src/components/Thumbnail/Item.tsx
```diff
@@ -13,34 +13,141 @@ import {
...
+                                icon: <IconArrowCircleUp20 />,
+                                onClick: onOpenFileDialog,
+                            },
+                            {
+                                title: 'Replace with asset',
+                                icon: <IconImageStack20 />,
+                                onClick: onAssetChooserClick,
+                            },
+                        ],
+                    ]}
```
celine-s, line 86: add filter here `filter((item) => item.length > 0)}` otherwise it looks now like this:

image

packages/thumbnail-grid-block/src/ThumbnailGridBlock.spec.ct.tsx
```diff
@@ -330,7 +264,97 @@ describe('Thumbnail Grid Block', () => {
         );
         cy.get(ThumbnailCaption).should('have.length', 3);
         cy.get(ThumbnailCaption).first().click();
-        cy.get(BlockItemWrapperBtn).first().click();
+        cy.get(BlockItemWrapperBtn).eq(1).click();
         cy.get(ThumbnailCaption).should('have.length', 2);
     });
+
+    it('should move the first item to fifth position by keyboard', () => {
```
celine-s, line 271: nice tests!

packages/animation-curve-block/src/components/Card.tsx
celine-s: same for other blocks. How about to move the tooltips into the shared component?

packages/dos-donts-block/src/DoDontItem.tsx
celine-s, line 1: Long titles are not supported within normal view but in drag vie they are...

normal: 
image

dragging:
image

packages/shared/src/components/BlockItemWrapper/Toolbar.tsx
celine-s, line 1: I am only able to focus the first item within all blocks. Could that be an issue of the content manager or does it work on your side?

celine-s (commented):
Nice work! It is really fun to use the drag n drop via keyboard! 😍 

Added some comments related to other stuff which probably accidentally happened via refactoring

packages/shared/src/components/BlockItemWrapper/Toolbar.tsx
ragi96: Same on my side with the new content manager, created a [task](https://app.clickup.com/t/861mzv61v) for checking what the issue is

packages/thumbnail-grid-block/src/ThumbnailGridBlock.tsx
```diff
@@ -156,7 +148,6 @@ export const ThumbnailGridBlock = ({ appBridge }: BlockProps) => {
     const thumbnailProps = {
         isEditing,
         thumbnailStyles,
-        showGrabHandle: isEditing && itemsState.length > 2,
```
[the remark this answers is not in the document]
ragi96: Sorry was removed during refactoring, and missed to re-add it again after

packages/thumbnail-grid-block/src/components/Thumbnail/Item.tsx
ragi96: Sorry was removed during refactoring, and missed to re-add it again after

packages/thumbnail-grid-block/src/ThumbnailGridBlock.tsx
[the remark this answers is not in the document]
celine-s: true, if the placeholder is not counted in that makes sense :D

packages/shared/src/components/BlockItemWrapper/Toolbar.tsx
celine-s: nice!

packages/dos-donts-block/src/DoDontItem.tsx
ragi96: There was a bug with the autoresize of that textarea, fixed that now multiline should be possible again edit/view and dragging mode

packages/animation-curve-block/src/components/Card.tsx
ragi96: moved them to a constants file, so they can be used everywhere

packages/animation-curve-block/src/components/Card.tsx
```diff
@@ -87,7 +88,11 @@ export const Card = forwardRef<HTMLDivElement, CardProps>(
                     shouldHideWrapper={replaceWithPlaceholder || !isEditing}
                     shouldHideComponent={replaceWithPlaceholder}
                     toolbarItems={[
-                        { icon: <IconArrowMove16 />, tooltip: 'Drag to move', draggableProps },
+                        {
+                            icon: <IconArrowMove16 />,
```
celine-s, line 92: you could even move this one to shared as it is everywhere the same Icon

packages/thumbnail-grid-block/src/ThumbnailGridBlock.tsx
celine-s, line 1: don't know if it did [jump](https://www.loom.com/share/77c9fbc2219f4fb59a8785aaa243b7f5) before or if I just didn't test any blocks in the middle of the page 🤔

celine-s: seems to be an issue within content manager.

celine-s (approved):

packages/animation-curve-block/src/components/Card.tsx
ragi96: won't do that as the icon and tooltip then could be optionally but it isn't in the other case

PR #596, opened 2023-06-19
feat(imageBlock, shared): add accessibility, extract set alt text flyout to shared and reuse in image block

celine-s:
[ticket](https://app.clickup.com/t/86782x7u0)


packages/shared/src/components/EditAltTextFlyout.tsx
ragi96, line 1: Test are missing, and would love to have it in its own folder as the other shared components

packages/animation-curve-block/src/AnimationCurveBlock.spec.ct.tsx
```diff
@@ -424,7 +424,6 @@ describe('AnimationCurve Block', () => {
         cy.get(FLYOUT_SELECTOR).should('exist');
```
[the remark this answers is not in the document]
celine-s: The test always failed. Imho it did not make sense; we move the end point to 0,50 butt afterwards test it and say it should equal 1 (line 431 +). Therefore, I thought either change the values on line 431 or remove the move endpoint. Do you know what I mean? Am I missing sth?

packages/shared/src/components/EditAltTextFlyout.tsx
celine-s: goood point! Will add 👍

packages/animation-curve-block/src/AnimationCurveBlock.spec.ct.tsx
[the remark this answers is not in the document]
ragi96: Then its all good

packages/shared/src/components/EditAltTextFlyout.spec.ct.tsx
ragi96, line 1: Can we move them into a folder "EditAltTextFlyout" so it has the same structure as the other shared components

packages/shared/src/components/EditAltTextFlyout.tsx
```diff
@@ -0,0 +1,96 @@
...
+    FlyoutFooter,
+    FlyoutPlacement,
+    FormControl,
+    HelperPosition,
+    IconCheckMark16,
+    TextInput,
+} from '@frontify/fondue';
+import React from 'react';
+
+type EditAltTextFlyoutProps = {
```
ragi96, line 17: Also move the types into a `types.ts` file within `EditAltTextFlyout` folder

ragi96 (commented):
Just some structure adjustments, so it has the same structure as the other components within `shared`

packages/shared/src/components/EditAltTextFlyout.spec.ct.tsx
celine-s: good point. Done 👍

packages/shared/src/components/EditAltTextFlyout.tsx
celine-s: 👍

ragi96 (approved):

PR #597, opened 2023-06-20
feat(shared): use onTextChange in the shared Rich text editor

celine-s:


packages/shared/src/components/RichTextEditor/types.ts
```diff
@@ -10,12 +10,9 @@ export type RichTextEditorProps = {
...
     gap?: string;
     plugins?: PluginComposer;
-    updateValueOnChange?: boolean;
     showSerializedText?: boolean;
-    shouldPreventPageLeave?: boolean;
-    onBlur?: (value: string) => void;
-    onTextChange?: (value: string) => void;
-    onValueChanged?: (value: TreeOfNodes | null) => void;
+    updateValueOnChange?: boolean;
+    onTextChange: (value: string) => void;
```
Walkeezy, line 15: Should this really be the only non-optional prop?

packages/text-block/src/TextBlock.tsx
```diff
@@ -36,10 +25,7 @@ export const TextBlock = ({ appBridge }: BlockProps): ReactElement => {
             gap={gap}
             plugins={getPlugins(appBridge, columnNumber, gap)}
             placeholder={PLACEHOLDER}
-            onBlur={onTextChange}
-            onTextChange={onTextChange}
-            onValueChanged={() => setIsApiRequestPending(true)}
-            shouldPreventPageLeave={isApiRequestPending}
+            onTextChange={(value: string) => setBlockSettings({ content: value })}
```
ragi96, line 28:
```suggestion
            onTextChange={(content: string) => setBlockSettings({ content })}
```

packages/shared/src/components/RichTextEditor/types.ts
celine-s: I see :D. I was just wondering if it ever can be undefined and I guessed "no". Didn't overthink the other ones, but know thinking about it maybe the id should be non-optional, but all others can be optional. Or what do you think?

packages/text-block/src/TextBlock.tsx
celine-s: good catch! Changed 👍

packages/asset-kit-block/src/AssetKitBlock.tsx
```diff
@@ -131,7 +131,7 @@ export const AssetKitBlock = ({ appBridge }: BlockProps): ReactElement => {
                             }
                             isEditing={isEditing}
                             plugins={new PluginComposer({ noToolbar: true }).setPlugin()}
-                            onBlur={(value) => setBlockSettings({ buttonText: value })}
+                            onTextChange={(value) => setBlockSettings({ buttonText: value })}
```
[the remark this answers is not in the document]
celine-s: 👍

packages/shared/src/components/RichTextEditor/types.ts
Walkeezy: I think we could keep `onTextChange` optional too, then we would not need to adjust all the tests.

celine-s: Changed 👍

ragi96 (commented):
Let's wait for the final feedback of Schirmi

fulopdaniel (approved):

PR #603, opened 2023-06-29
feat(ImageBlock): fix ux review

celine-s:
tickets: [image block general](https://app.clickup.com/t/86783gd9e) and [accessibility](https://app.clickup.com/t/86782x7u0)

* Attachment button same size as download button
* Buttons now always stays on top of image (despite it's padding or image size)
* Focusing order corrected

### DIFF
--- a/packages/image-block/src/components/helpers.ts
+++ b/packages/image-block/src/components/helpers.ts
@@ -74,6 +74,19 @@ export const getCaptionPlugins = (appBridge: AppBridgeBlock) =>
             new ResetFormattingPlugin(),
         ]);
 
+export const getTotalImagePadding = (blockSettings: Settings): CSSProperties => {
+    const border = blockSettings.hasBorder ? blockSettings.borderWidth?.replace('px', '') : undefined;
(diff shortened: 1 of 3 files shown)


fulopdaniel (approved):

packages/image-block/src/components/helpers.ts
```diff
@@ -74,6 +74,19 @@ export const getCaptionPlugins = (appBridge: AppBridgeBlock) =>
             new ResetFormattingPlugin(),
         ]);
 
+export const getTotalImagePadding = (blockSettings: Settings): CSSProperties => {
```
ragi96, line 77: Add a unit test for that function

celine-s: We don't have that package within this block. I just added 3 tests to the ImageBlock.spec.ct.tsx (and fixed already a bug :D)

ragi96: For me this is also fine

ragi96 (approved):
Thanks

PR #607, opened 2023-07-03
fix(DoAndDont): heading order accessibility

ragi96:
[Task](https://app.clickup.com/t/85ywx2ec0)

### DIFF
--- a/packages/dos-donts-block/src/DoDontItem.tsx
+++ b/packages/dos-donts-block/src/DoDontItem.tsx
@@ -263,7 +263,8 @@ export const DoDontItem = React.forwardRef<HTMLDivElement, DoDontItemProps>(
                             </div>
                         )}
                         <div className="tw-w-full tw-flex tw-items-center">
-                            <h3
+                            <span
+                                className="a-h3"
                                 style={{
                                     marginBottom: 0,
                                     marginTop: 0,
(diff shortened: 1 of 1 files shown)


packages/dos-donts-block/src/DoDontItem.tsx
```diff
@@ -263,7 +263,8 @@ export const DoDontItem = React.forwardRef<HTMLDivElement, DoDontItemProps>(
                             </div>
                         )}
                         <div className="tw-w-full tw-flex tw-items-center">
-                            <h3
+                            <span
+                                className="a-h3"
```
fulopdaniel, line 267: Can we use the theme settings here maybe to get the heading3 styles?

ragi96: They are already used, see the lines below... 

The class is just needed if the font-family is inherited

fulopdaniel (approved):

packages/dos-donts-block/src/DoDontItem.tsx
```diff
@@ -263,7 +263,8 @@ export const DoDontItem = React.forwardRef<HTMLDivElement, DoDontItemProps>(
                             </div>
                         )}
                         <div className="tw-w-full tw-flex tw-items-center">
-                            <h3
+                            <span
```
utmishra, line 266: Just checking, maybe we should add `aria-label` and `title` for accessibility?

utmishra (approved):
Left a comment but can be approved already

packages/dos-donts-block/src/DoDontItem.tsx
ragi96: Wouldn't use title as it would just help mouse users with a tooltip. 

I don't think this is needed as the shown content would be the same as the `aria-label`, or what do you have in mind to add there? That would be helpful for disabled people. 

[More infos](https://www.aditus.io/aria/aria-label/#over-using-it)

PR #608, opened 2023-07-04
feat(imageBlock): deactivate link

ragi96:
As discussed with Schirmi and Luca, we will release the Image Block (Beta), without the link chooser and update the block as soon the new link chooser is available


utmishra (approved):
Verified impact of removal. Looks safe 👍🏽

fulopdaniel (approved):

PR #617, opened 2023-07-18
fix: update toolbar focus styles

SamCreasey:


fulopdaniel (approved):

packages/shared/src/components/BlockItemWrapper/Toolbar.tsx
```diff
@@ -3,11 +3,13 @@
 import React from 'react';
 import {
     ActionMenu,
+    FOCUS_VISIBLE_STYLE,
     Flyout,
     IconDotsHorizontal16,
     MenuItemContentSize,
     Tooltip,
     TooltipPosition,
+    merge,
```
ragi96, line 12: Why adding merge? Can you please use the `joinClassNames` or replace all of them, as its basically the same

SamCreasey: fixed

ragi96 (approved):
Thanks

PR #618, opened 2023-07-19
fix(ThumbnailGrid): mass image upload not working

fulopdaniel:
Fix for https://frontify.slack.com/archives/C04533CM2BU/p1689696811622299


ragi96 (approved):
Approved but i opened a new task, so we refactoring the whole add topic as I think we can do that better 

https://app.clickup.com/t/861n2zve1

PR #634, opened 2023-08-25
feat: use new useAssetBulkDownload hook from AppBridge

hochreutenerl:
- Replacing the old hook with the new one
- Allow all assets types via the assetChooser
- fix a bug in which an asset could be replaced with an already used asset, what could end up in replacing 2 with 1 if you would have done that

[Task BulkDownload](https://app.clickup.com/t/861n31u4y)
[Bug all asset types](https://app.clickup.com/t/2523021/TASK-8540)


packages/asset-kit-block/src/components/AssetGrid.tsx
```diff
@@ -37,7 +38,8 @@ export const AssetGrid = ({
                     {currentAssets
                         ? currentAssets.map((asset) => (
                               <ThumbnailItem
-                                  key={asset.id}
+                                  key={`${asset.id}-${generateRandomId()}`}
```
fulopdaniel, line 41: Why is this needed? As I think this will generate a new ID each rerender => it will remount the Thumbnail item each rerender which might not be what you want. Could you use the blockId here or the [React useID() hook](https://react.dev/reference/react/useId) which creates IDs that persist across rerenders?

packages/asset-kit-block/src/components/AssetSelection.tsx
```diff
@@ -24,17 +24,16 @@ export const AssetSelection = ({
 
     const onOpenAssetChooser = () => {
         openAssetChooser(
-            (assetsObject) => {
+            async (assetsObject) => {
                 setIsUploadingAssets(true);
                 const assetsIds = Array.from(assetsObject).map((asset) => asset.id);
                 saveDownloadUrl('');
-                addAssetIdsToKey(ASSET_SETTINGS_ID, assetsIds).then(() => setIsUploadingAssets(false));
+                await addAssetIdsToKey(ASSET_SETTINGS_ID, assetsIds).then(() => setIsUploadingAssets(false));
```
fulopdaniel, line 31: Will this keep the asset chooser open until the API request is done? What was the  issue with showing the loading indicator?

packages/asset-kit-block/src/components/AssetGrid.tsx
ragi96: Good catch its not needed anymore, did first that and then saw what the real problem is, that the same asset could be added multiple times

packages/asset-kit-block/src/components/AssetSelection.tsx
ragi96: Reverted, as it was highly useless in combination with `.then`

fulopdaniel (approved):

PR #648, opened 2023-09-06
fix(DoDont): safari webkit-text-fill-color

ragi96:
Will fix the discrepancy of the color in safari between view and edit mode


fulopdaniel (approved):

packages/dos-donts-block/src/DoDontItem.tsx
```diff
@@ -277,7 +277,6 @@ export const DoDontItem = React.forwardRef<HTMLDivElement, DoDontItemProps>(
                         )}
                         <div className="tw-w-full tw-flex tw-items-center">
                             <span
-                                className="a-h3"
```
fulopdaniel, line 280: Is this not needed anymore?

ragi96: not needed anymore, as it was just used for `font-family: inherit` what got refactored and can't be the case anymore

PR #649, opened 2023-09-06
chore: update node, cypress and prettier

SamuelAlev:
To test (manual change happened in these blocks):
- Callout `none` icon (replaced `Fragment` with `null`)
- Color Kit (replaced `Fragment` with `null`)
- Color (replaced `Fragment` with `null`)
- Image (removed extra `Fragment`)
- Quote `none` icon (replaced `Fragment` with `null`)
- Storybook  (removed extra `Fragment`)
- Thumbnail Grid (replaced `Fragment` with `null`)


examples/asset-upload/package.json
```diff
@@ -13,18 +13,18 @@
         "typecheck": "tsc --noEmit"
     },
     "devDependencies": {
-        "@babel/core": "^7.22.11",
-        "@frontify/eslint-config-react": "^0.15.5",
+        "@babel/core": "^7.22.15",
+        "@frontify/eslint-config-react": "0.16.1",
```
ragi96, line 17:
```suggestion
        "@frontify/eslint-config-react": "0.16.2",
```

Or is there a problem with that version?

SamuelAlev: Yup, I need to fix an issue with the `prefer-async-to-then` lint rule (Cypress use `.then` but not in a promise context)

PR #660, opened 2023-09-14
chore: rename OrderableList component to LegacyOrderableList

anxobotana:
In order to get rid of `react-dnd` in fondue `OrderableList` component has been renamed to `LegacyOrderableList` and deprecated. 
The only component using this component here is `CheckList` so it has been updated accordingly.

### DIFF
--- a/packages/checklist-block/src/types.ts
+++ b/packages/checklist-block/src/types.ts
@@ -1,6 +1,6 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
-import { ButtonSize, Color, LegacyItemDragState, OrderableListItem } from '@frontify/fondue';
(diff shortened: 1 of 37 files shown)


packages/checklist-block/src/types.ts
```diff
@@ -1,13 +1,13 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
-import { ButtonSize, Color, ItemDragState, OrderableListItem } from '@frontify/fondue';
+import { ButtonSize, Color, LegacyItemDragState, OrderableListItem } from '@frontify/fondue';
```
anxobotana, line 3: Probably you want to use `LegacyOrderableListItem` as well

ragi96: you are right, adjusted all of them

fulopdaniel (approved):

PR #674, opened 2023-10-02
fix(ImageBlock): image can be too narrow

fulopdaniel:
If the devicePixelRatio was around 0.9 then the image was not full width.


ragi96 (approved):
👍

PR #675, opened 2023-10-02
fix(CalloutBlock): change colors based on HubAppearance

ragi96:
Adds a listener on `HubAppearanceUpdate` so it adjusts the colors without reload, if the accent-color got changed

- Also refactored the tests

[Bug](https://app.clickup.com/t/862k5d8wb)

Needs: https://github.com/Frontify/web-app/pull/1370


packages/callout-block/src/CalloutBlock.tsx
```diff
@@ -35,23 +59,28 @@ export const CalloutBlock = ({ appBridge }: BlockProps): ReactElement => {
...
+        const styleElement = document.getElementById('design-settings');
+        if (!styleElement) {
+            return;
         }
-    };
 
-    const accentColor = getAccentColor(blockSettings.type);
-    const backgroundColor = blockSettings.appearance === Appearance.Strong ? accentColor : setAlpha(0.1, accentColor);
-    const textColor = getTextColor(blockSettings.appearance, accentColor, backgroundColor);
+        const observer = new MutationObserver(() => {
```
SamuelAlev, line 74: Instead of observer could you use the CSS variable and add a local CSS variable?

```html
<div id="upper" style="--my-var: blue;">
  <div id="upper" style="--my-var: red;"></div> <!-- become red -->
</div>
```

ragi96: Could do that, but the observer would still be needed, as the local variable would need to get recalculated after the change within the global theme settings.

SamuelAlev: Optimally you shouldn't need the local variable, why is it needed?

ragi96: Because based on the setting `appearance` (Light/Dark), the background color gets calculated and also the text color. Which is based on the type (Info/Note/Warning/Tip), which then is the accent color that should be taken as base for the calculation.

fulopdaniel: Instead of observing the DOM, could you listen to the `HubAppearanceUpdated` event that is emitted with window.emitter?

packages/callout-block/src/CalloutBlock.tsx
```diff
@@ -12,16 +12,40 @@ import {
...
+        case Type.Note:
+            return style.getPropertyValue(`${THEME_PREFIX}accent-color-note-color`);
+        case Type.Tip:
+            return style.getPropertyValue(`${THEME_PREFIX}accent-color-tip-color`);
+        case Type.Warning:
+            return style.getPropertyValue(`${THEME_PREFIX}accent-color-warning-color`);
+    }
+};
+
+const computeStyles = (type: Type, appearance: Appearance) => {
```
fulopdaniel, line 37: Can functions outside of the component move to helpers folder, similar to getTextColor? Maybe create a color.ts file that holds all the color related helper functions.

packages/callout-block/src/CalloutBlock.tsx
ragi96: That's most likely the solution, thanks :)

packages/callout-block/src/CalloutBlock.tsx
```diff
@@ -35,23 +37,19 @@ export const CalloutBlock = ({ appBridge }: BlockProps): ReactElement => {
...
-        }
-    };
+    useEffect(() => {
+        const updateStyles = () => {
+            const { backgroundColor, textColor } = computeStyles(type, appearance);
+            setBackgroundColor(backgroundColor);
+            setTextColor(textColor);
+        };
+
+        window.emitter.on('HubAppearanceUpdated', () => {
```
fulopdaniel, line 47: Will this create a new subscription every time this useEffect runs? I suggest to unsubscribe in the return function of this useEffect.

ragi96: You are completely right, missed that one

fulopdaniel (approved):

PR #691, opened 2023-11-21
fix: Template block feedback bug fixes [CU-8692wq8w6]

ruslan-sazonov:
This PR introduces the template block's `title` and `description` related code cleanup and button justification fix.


packages/template-block/src/components/TemplateText.tsx
```diff
@@ -34,33 +45,38 @@ export const TemplateText = ({
...
+            id="template-block-title"
+            value={title ?? convertToRteValue(TextStyles.heading3)}
+            placeholder="Add a title"
+            onTextChange={setTitle}
+            isEditing={isEditing}
+            plugins={customTitlePlugins}
+        />
+    );
+
+    const memoDescriptionRte = (
```
fulopdaniel, line 67: These are no longer memoized, be careful as it can easily get laggy. If you decided that it was not needed then remove the memo from name and just use it directly.

packages/template-block/src/components/TemplateText.tsx
```diff
@@ -34,33 +45,38 @@ export const TemplateText = ({
...
+            .setPlugin([new BoldPlugin(), new ItalicPlugin(), new UnderlinePlugin(), new StrikethroughPlugin()])
+            .setPlugin([new ResetFormattingPlugin(), new AutoformatPlugin()]);
+    }, []);
+
+    const memoTitleRte = (
+        <RichTextEditor
+            id="template-block-title"
+            value={title ?? convertToRteValue(TextStyles.heading3)}
+            placeholder="Add a title"
+            onTextChange={setTitle}
```
fulopdaniel, line 61: just a note: this will call the API every time onTextChange emits. There is some built-in debounce so it is probably not an issue, but something to be mindful of.

fulopdaniel (approved):

packages/template-block/src/components/TemplateText.tsx
ruslan-sazonov: True. I've been experimenting in order to fix the RTE updates while being in edit mode.

packages/template-block/src/TemplateBlock.tsx
```diff
@@ -107,6 +100,18 @@ export const TemplateBlock = ({ appBridge }: BlockProps): ReactElement => {
...
 
+    const saveDescription = (newDescription: string) => {
+        if (description !== newDescription) {
+            updateBlockSettings({ description: newDescription });
+        }
+    };
+
+    const saveTitle = (newTitle: string) => {
+        if (title !== newTitle) {
+            updateBlockSettings({ title: convertToRteValue(TextStyles.heading3, newTitle) });
```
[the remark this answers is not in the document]
ruslan-sazonov: Yep, the template replacement breaks the initial `heading3` style for a title.

packages/template-block/src/components/TemplateText.tsx
ruslan-sazonov: Got it. Thanks for noticing 👍

PR #693, opened 2023-11-28
feat(ImageBlock): allow types other than images

fulopdaniel:
https://app.clickup.com/t/8678far4v

### DIFF
--- a/packages/image-block/src/ImageBlock.tsx
+++ b/packages/image-block/src/ImageBlock.tsx
@@ -2,8 +2,6 @@
 
 import {
     Asset,
-    AssetChooserObjectType,
-    FileExtensionSets,
     useAssetChooser,
     useAssetUpload,
     useBlockAssets,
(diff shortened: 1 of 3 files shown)


ragi96 (changes requested):
Not working as expected, as its still restricted.

For direct upload you need to adjust more within `ImageBlock.tsx`

Line 50: `const [openFileDialog, { selectedFiles }] = useFileInput({});`

And adjust the `objectTypes` and `extensions` for the AssetChooser Line 75/76

ragi96 (commented):
Screenshot 2023-11-29 at 11 22 05

For audio and video files which can now be selected the icon is broken as its using the original file type, txt, zip, docx works as expected.

I think you should use the same settings as within the old image block, then it should be fine. And also adjust it within the settings.ts

ragi96 (approved):

PR #696, opened 2023-12-06
fix(GradientBlock): opacity is not working

fulopdaniel:
1. Fix hex opacity not working in gradient stops. (https://app.clickup.com/t/2523021/TASK-9383)
2. Fix z-index issue when hovering color. (https://app.clickup.com/t/86939b2be)


SamCreasey (approved):
Z index fix works well 👍

PR #701, opened 2024-01-04
fix: change default audio positioning

SamCreasey:
Default positioning has been updated to reflect the positioning of the old audio block.
The title and description should be by default above the audio block, until the user changes this value.


silviojaeger (approved):
Tutti bene✅

PR #703, opened 2024-01-09
feat(*): add css scoping to each block

fulopdaniel:
Created a PostCSS Plugin that prepends every css rule in the generated style.css with an extra selector, essentially scoping the whole css file to the specific block.

```
.foo .bar { ... }
```
becomes
```
.text-block .foo .bar { ... }
```

Previously we relied on Tailwind's `important` scoping approach, but that does not scope css rules that were imported to the package (e.g. @frontify/guideline-blocks-settings/dist/style.css)

To test:
1. run `pnpm run deploy --dryRun` inside any package
2. check the style.css inside the /dist folder of the package
3. all rules should be scoped

### DIFF
--- a/postcss/scope.js
+++ b/postcss/scope.js
@@ -20,15 +20,16 @@ module.exports = (opts = {}) => {
                 ) {
                     return;
                 }
-                rule.selectors = rule.selectors.map((selectors) => {
-                    return selectors.split(/,\s* /g).map((selector) => {
-                        if (selector === opts.selector) {
-                            return selector;
-                        }
-                        const newSelector = `${opts.scope} ${selector}`;
-                        return newSelector;
-                    });
-                });
+                rule.selectors = rule.selectors.map((originalSelector) =>
+                    originalSelector
+                        .split(/,\s*/g)
+                        .map((individualSelector) =>
+                            individualSelector === opts.selector
+                                ? individualSelector
+                                : `${opts.scope} ${individualSelector}`,
+                        )
+                        .join(", "),
+                );
             });
         },
     };
--- a/postcss/scope.js
+++ b/postcss/scope.js
@@ -1,0 +1,37 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+/**
+ * This custom plugin prepends every css rule in the generated style.css
+ * with an extra selector, essentially scoping the whole css file to the specific block
+ */
+
+/**
+ * @type {import('postcss').PluginCreator}
+ */
+module.exports = (opts = {}) => {
+    return {
+        postcssPlugin: "scope",
+        Root(root) {
+            root.walkRules((rule) => {
+                if (
+                    rule.parent &&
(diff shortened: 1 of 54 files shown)


fulopdaniel:
https://app.clickup.com/t/8693f384y

postcss/scope.js
```diff
@@ -0,0 +1,37 @@
...
+            root.walkRules((rule) => {
+                if (
+                    rule.parent &&
+                    rule.parent.type === "atrule" &&
+                    rule.parent.name.includes("keyframes")
+                ) {
+                    return;
+                }
+                rule.selectors = rule.selectors.map((selectors) => {
+                    return selectors.split(/,\s* /g).map((selector) => {
```
ragi96, line 24: Why the space in the regex?

postcss/scope.js
```diff
@@ -0,0 +1,37 @@
...
+                        if (selector === opts.selector) {
+                            return selector;
+                        }
+                        const newSelector = `${opts.scope} ${selector}`;
+                        return newSelector;
+                    });
+                });
```
ragi96, line 31:
```suggestion
               rule.selectors = rule.selectors.map(originalSelector =>
                    originalSelector.split(/,\s*/g).map(individualSelector => {
                        return individualSelector === opts.selector ? individualSelector : `${opts.scope} ${individualSelector}`;
                    }).join(", ") // Joining back individual selectors into a combined form
                );
```

Comment is just for clarification

hochreutenerl:
Nice direction where this is going. @fulopdaniel 
Would it even be possible to scope with the appId (read it from the manifest), so we can have a solution that also works for external blocks?

ragi96 (approved):
And I noticed that the css build now is smaller and I don't get why thats the case as it should be bigger, are we missing something?

fulopdaniel:
@hochreutenerl We could read the manifest.json, but that does not automatically solve the issue for external blocks. They will need to configure their postcss.config.js themselves either way.

@ragi96 I will look into the point with the css size. One potential reason why the bundle could be smaller is that I removed the tailwind important part, so that is no longer part of the bundle.

postcss/scope.js
fulopdaniel: It was actually automatically joining the individual selectors in the end, but I added the join(", ") anyway to make sure. Thanks!

postcss/scope.js
fulopdaniel: It's a mistake, thanks!

ragi96 (approved):
LGTM

hochreutenerl:
@fulopdaniel I think we can also adjust our [templates](https://github.com/Frontify/brand-sdk/tree/main/packages/cli/templates) - not all of them use postcss though - and add a matching class to wrapping div.
It's fine for me to do it as a later stage.

ragi96:
@hochreutenerl then we probably should release the plugin properly so we don't have to copy paste the code

PR #704, opened 2024-01-11
fix(*): css scoping regex is incorrect

fulopdaniel:
1. The CSS scoping postCSS plugin regex was wrong. Tailwind classes like `bg-[rgb(150,150,150)]` were wrongly prepended because of the commas. Updated the regex with a negative lookahead, to not prepend with the selector if the comma is escaped. (tailwind transforms classes like that to `bg-[rgb(150\,150\,150)]`).


ragi96 (approved):
LGTM

PR #708, opened 2024-01-17
refactor: move template block data into hook [CU-8693jhuuw]

mikeyrayvon:
I'm working on fixes for the Template Block layout issues.

This PR just moves the Template Block data handling into a custom hook, to make the code more legible and easier to bugfix/refactor

Fixes https://app.clickup.com/t/8693jhuuw

### DIFF
--- a/packages/template-block/src/hooks/useTemplateBlockData.ts
+++ b/packages/template-block/src/hooks/useTemplateBlockData.ts
@@ -1,0 +1,147 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import {
+    Template,
+    TemplateLegacy,
+    closeTemplateChooser,
+    useBlockAssets,
+    useBlockSettings,
+    useBlockTemplates,
+    useEditorState,
+} from '@frontify/app-bridge';
+import { TextStyles, generateRandomId } from '@frontify/fondue';
+import { useCallback, useEffect, useState } from 'react';
+import { PreviewType, Settings, TextPositioningType, textPositioningToStyles } from '../types';
+import { TEMPLATE_BLOCK_SETTING_ID } from '../constants';
+import { BlockProps, convertToRteValue, radiusStyleMap, toRgbaString } from '@frontify/guideline-blocks-settings';
+import { getIsRows } from '../helpers/layout';
+
+export const useTemplateBlockData = (appBridge: BlockProps['appBridge']) => {
+    const [blockSettings, updateBlockSettings] = useBlockSettings<Settings>(appBridge);
+    const [selectedTemplate, setSelectedTemplate] = useState<Template | null>(null);
+    const [lastErrorMessage, setLastErrorMessage] = useState('');
+    const [templateTextKey, setTemplateTextKey] = useState(generateRandomId);
+    const isEditing = useEditorState(appBridge);
+    const { blockAssets } = useBlockAssets(appBridge);
+    const { blockTemplates, updateTemplateIdsFromKey, error } = useBlockTemplates(appBridge);
+
+    const {
+        title,
+        description,
+        preview,
+        hasBackground,
+        backgroundColor,
+        hasBorder_blockCard,
+        borderWidth_blockCard,
+        borderColor_blockCard,
+        borderStyle_blockCard,
+        hasRadius_blockCard,
+        radiusValue_blockCard,
+        radiusChoice_blockCard,
+        textPositioning,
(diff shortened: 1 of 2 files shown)


packages/template-block/src/hooks/useTemplateBlockData.ts
```diff
@@ -0,0 +1,147 @@
...
+    closeTemplateChooser,
+    useBlockAssets,
+    useBlockSettings,
+    useBlockTemplates,
+    useEditorState,
+} from '@frontify/app-bridge';
+import { TextStyles, generateRandomId } from '@frontify/fondue';
+import { useCallback, useEffect, useState } from 'react';
+import { PreviewType, Settings, TextPositioningType, textPositioningToStyles } from '../types';
+import { TEMPLATE_BLOCK_SETTING_ID } from '../constants';
```
SamCreasey, line 15: Please re-order imports to match web-app. Although we don't have eslint enforced here it adds to readability.

Also any type imports should be explicitly set as type (eg: import { type ReactElement } from 'react')

packages/template-block/src/hooks/useTemplateBlockData.ts
```diff
@@ -0,0 +1,147 @@
...
+
+    useEffect(() => {
+        if (error !== null) {
+            setLastErrorMessage(error);
+        }
+    }, [error]);
+
+    useEffect(() => {
+        if (blockTemplates[TEMPLATE_BLOCK_SETTING_ID] && blockTemplates[TEMPLATE_BLOCK_SETTING_ID].length > 0) {
+            const lastTemplate = blockTemplates[TEMPLATE_BLOCK_SETTING_ID].pop();
```
SamCreasey, line 114: Avoid mutating method here

packages/template-block/src/hooks/useTemplateBlockData.ts
```diff
@@ -0,0 +1,147 @@
...
+import { PreviewType, Settings, TextPositioningType, textPositioningToStyles } from '../types';
+import { TEMPLATE_BLOCK_SETTING_ID } from '../constants';
+import { BlockProps, convertToRteValue, radiusStyleMap, toRgbaString } from '@frontify/guideline-blocks-settings';
+import { getIsRows } from '../helpers/layout';
+
+export const useTemplateBlockData = (appBridge: BlockProps['appBridge']) => {
+    const [blockSettings, updateBlockSettings] = useBlockSettings<Settings>(appBridge);
+    const [selectedTemplate, setSelectedTemplate] = useState<Template | null>(null);
+    const [lastErrorMessage, setLastErrorMessage] = useState('');
+    const [templateTextKey, setTemplateTextKey] = useState(generateRandomId);
```
SamCreasey, line 23:
```suggestion
    const [templateTextKey, setTemplateTextKey] = useState(generateRandomId()); 
```

packages/template-block/src/hooks/useTemplateBlockData.ts
```diff
@@ -0,0 +1,147 @@
...
+        if (blockTemplates[TEMPLATE_BLOCK_SETTING_ID] && blockTemplates[TEMPLATE_BLOCK_SETTING_ID].length > 0) {
+            const lastTemplate = blockTemplates[TEMPLATE_BLOCK_SETTING_ID].pop();
+
+            if (lastTemplate) {
+                setSelectedTemplate(lastTemplate);
+            }
+        }
+    }, [blockTemplates]);
+
+    return {
```
SamCreasey, line 122: Just a suggestion but I wonder if it makes sense with this many props to split these into relevant sub-objects. Also does the end user need all data such as "preview, previewCustom and hasPreview"? Should these properties be resolved into a single consumable value before returning?

SamCreasey (changes requested):

mikeyrayvon:
Hey thanks for your comments @SamCreasey ! 💯 

To clarify, I'm working on a lot of fixes and cleaning of this block. This PR was strictly to move the code over to make it easier to clean. 

I'll make your suggested improvements here, and I'm sure there will be more to come in the following PRs 👍🏻

packages/template-block/src/hooks/useTemplateBlockData.ts
mikeyrayvon: Yea I agree and considered this as well. I will try and improve how the data is organized in later branches

SamCreasey (approved):
Looks good to me :)

PR #709, opened 2024-01-17
feat: Added page count toggle [CU-8693hv558]

findmind:
Closes https://app.clickup.com/t/8693hv558

This PR adds a page count toggle to show or hide the page count of a template.

Todo:

- [x] Test page count toggle
- [ ] Merge main & fix tests

Screenshot of the setting disabled; by default it's enabled and the page count will be shown between the description & title:
image


ryancarville (approved):
Tested and LGTM 🥳  Nice work!

One call out (possibly out of scope for this PR):  on initial rendering of the block, the `flex-direction` will switch from `column` to `row`.  Once a template is loaded, the `flex-direction` sticks with `column`

findmind:
> Tested and LGTM 🥳 Nice work!
> 
> One call out (possibly out of scope for this PR): on initial rendering of the block, the `flex-direction` will switch from `column` to `row`. Once a template is loaded, the `flex-direction` sticks with `column`

Thanks! Yeah, I'm looking at the layout issues in a different PR; just found out that there are some tailwind prefixing issues, so need to get that resolved first.

packages/template-block/src/TemplateBlock.tsx
```diff
@@ -195,9 +195,12 @@ export const TemplateBlock = ({ appBridge }: BlockProps): ReactElement => {
                                 <TemplateText
                                     appBridge={appBridge}
                                     title={title}
-                                    blockSettings={blockSettings}
                                     description={description}
-                                    pageCount={selectedTemplate?.pages.length ?? 0}
+                                    pageCount={
+                                        blockSettings.hasPageCount !== false
+                                            ? selectedTemplate?.pages.length ?? 0
+                                            : undefined
```
SamCreasey, line 202:
```suggestion
                                        blockSettings.hasPageCount
                                            ? selectedTemplate?.pages.length ?? 0
                                            : undefined
```

Minor but hasPageCount should have it's default value populated when the block is created, so it doesn't need to consider `true | undefined` because it should only be able to be boolean.

SamCreasey (approved):
Approved with 1 small comment but it is more of a personal preference thing I feel

packages/template-block/src/TemplateBlock.tsx
findmind: Thanks for the hint! 🚀

PR #711, opened 2024-01-18
feat(CLI): add command to deploy all blocks

fulopdaniel:
`sh cli/deploy_all_blocks.sh`


ragi96:
Will close this one, we could do that now via CI, if we wanna do it

PR #713, opened 2024-01-18
fix: Template block layout issues [CU-8693jga13]

findmind:
Fixes https://app.clickup.com/t/8693jga13

- Left/right layout not working -> Fixed by changing import order, fixed issues with applying TW classes in the right order
- Improved gaps (removed unnecessary vertical gaps)
- Cleaned up a bit

The highlighted settings in the screenshot should now work properly, along with all the other layout options. Vertical gaps and spacing should also be improved.

image


packages/template-block/src/helpers/layout.ts
```diff
@@ -20,14 +20,10 @@ export const getCardPadding = (blockSettings: Settings) => {
 
 export const getIsRows = (hasPreview: boolean, textPositioning: TextPositioningType): boolean => {
     const flexDirection = hasPreview ? textPositioningToFlexDirection[textPositioning] : 'row';
-
-    return hasPreview && (flexDirection === 'row' || flexDirection === 'row-reverse');
+    return hasPreview && ['row', 'row-reverse'].includes(flexDirection);
 };
```
ragi96, line 24:
```typescript
export const getIsRows = (hasPreview: boolean, textPositioning: TextPositioningType): boolean => {
    return hasPreview && ['row', 'row-reverse'].includes(textPositioningToFlexDirection[textPositioning]);
};
``` 

How about that?

packages/template-block/src/helpers/layout.ts
```diff
@@ -20,14 +20,10 @@ export const getCardPadding = (blockSettings: Settings) => {
...
 
 export const getLayoutClasses = (hasPreview: boolean, textPositioning: TextPositioningType) => {
-    const previewClasses = getIsRows(hasPreview, textPositioning)
-        ? 'tw-flex-col'
-        : 'tw-grid tw-grid-rows-2 grid-flow-col';
-
+    const previewClasses = getIsRows(hasPreview, textPositioning) ? 'tw-flex-col' : 'tw-grid grid-flow-col';
     return hasPreview ? previewClasses : 'tw-grid tw-grid-cols-3';
 };
```
ragi96, line 29:
```typescript
export const getLayoutClasses = (hasPreview: boolean, textPositioning: TextPositioningType): string => {
    if (!hasPreview) {
        return 'tw-grid tw-grid-cols-3';
    }
    return getIsRows(hasPreview, textPositioning) ? 'tw-flex-col' : 'tw-grid grid-flow-col';
};
``` 

Nit: but imo improves readability

ryancarville (approved):
Tested and working on my end!  Nice work! 👍

packages/template-block/src/helpers/layout.ts
findmind: Well spotted, this is a very weird double-condition for some reason.

ragi96 (approved):
LGTM

PR #717, opened 2024-01-19
fix(AudioBlock): add title automatically if title is empty

ragi96:
- fixes the issue with not adding title if title is empty and a audio asset gets added
- adds tests to avoid regression
- replaces deprecated getBlockId function

[Task](https://app.clickup.com/t/8693jjk0q)

### DIFF
--- a/packages/audio-block/src/AudioBlock.spec.ct.tsx
+++ b/packages/audio-block/src/AudioBlock.spec.ct.tsx
@@ -12,11 +12,13 @@ const AudioBlockSelector = '[data-test-id="audio-block"]';
 const AudioTagSelector = '[data-test-id="audio-block-audio-tag"]';
 const UploadPlaceholderSelector = '[data-test-id="block-inject-button"]';
 const AudioBlockTitleHtmlSelector = '[data-test-id="block-title"] [data-test-id="rte-content-html"]';
+const AudioBlockTitleRteSelector = '[data-test-id="block-title"]';
 const AudioBlockDescriptionHtmlSelector = '[data-test-id="block-description"] [data-test-id="rte-content-html"]';
 const DownloadButtonSelector = '[data-test-id="download-button"]';
 const AttachmentsTriggerSelector = '[data-test-id="attachments-flyout-button"]';
 const ViewModeAddonsSelector = '[data-test-id="view-mode-addons"]';
 const BlockWrapperSelector = '[data-test-id="block-item-wrapper"]';
+const UploadMenu = '[data-test-id="menu-item-title"]';
 
 const Title = '[{"type":"heading3","children":[{"text":"Audio Title"}]}]';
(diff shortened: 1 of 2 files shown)


packages/audio-block/src/AudioBlock.tsx
```diff
@@ -138,7 +141,8 @@ export const AudioBlock = withAttachmentsProvider(({ appBridge }: BlockProps) =>
                     <div className="tw-flex-1">
                         <div data-test-id="block-title">
                             <RichTextEditor
-                                id={`${appBridge.getBlockId().toString()}-title`}
+                                key={titleKey}
+                                id={`${appBridge.context('blockId').get()}-title`}
```
SamCreasey, line 145:
```suggestion
                                id={`${blockId}-title`}
```

Since we use it twice it would be nice to save it in a readable variable

packages/audio-block/src/AudioBlock.tsx
```diff
@@ -86,7 +89,7 @@ export const AudioBlock = withAttachmentsProvider(({ appBridge }: BlockProps) =>
         }
     };
 
-    const onTitleChange = (title: string) => title !== blockSettings.title && setBlockSettings({ title });
+    const onTitleChange = async (title: string) => title !== blockSettings.title && (await setBlockSettings({ title }));
```
SamCreasey, line 92:
```suggestion
    const onTitleChange = async (title: string) => {
    if(title !== blockSettings.title){
    await setBlockSettings({ title });
    }}
```
Otherwise the typing on the return of this function ends up being false | Promise<void> which feels a bit off

packages/audio-block/src/AudioBlock.spec.ct.tsx
```diff
@@ -219,4 +221,14 @@ describe('Audio Block', () => {
         mount(<AudioBlockWithStubs />);
         cy.get(ViewModeAddonsSelector).should('not.exist');
     });
+
+    it.only('should add the file name to the title when the title is empty', () => {
```
SamCreasey, line 225:
```suggestion
    it('should add the file name to the title when the title is empty', () => {
```

SamCreasey: Is this actually expected behaviour or should it always overwrite when the asset changes? Would be good to know if this has already been discussed with Martin/Schirmi

packages/audio-block/src/AudioBlock.tsx
ragi96: True didn't recognized it

packages/audio-block/src/AudioBlock.spec.ct.tsx
ragi96: I has been discussed last year and it should just be set when the asset gets added and no title is set. But yeah will add a second test for that

SamCreasey (approved):

PR #720, opened 2024-01-22
fix: Decouple card padding from background and border [CU-8693jg9px]

findmind:
This PR decouples the card padding from the background and border settings. Previously, the card padding was only applied if at least the background or border was enabled.

Fixes https://app.clickup.com/t/8693jg9px

image


utmishra (approved):
LGTM 👍🏽

PR #723, opened 2024-01-23
fix: Template Block align title and CTA center when only title and no preview is present [CU-8693kt73x]

findmind:
Fixes https://app.clickup.com/t/8693kt73x

- When only title is present (when not editing, preview hidden, page counter hidden, no description), title and CTA button should be vertically centered
- Remove margin top and bottom from title style to actually get the title in the middle
- Remove margin top on title always, so it's aligned with the rest of the block (see screenshot)

⚠️  2 caveats: 
- Since the vertical centering is dependent on `const hasTitleOnly = !hasPreview && !hasPageCount && !hasRichTextValue(description);`, the margins jump around when e.g. the pageCount setting or the description is changed.
- Since we overwrite the marginTop/marginBottom on all elements (rather than just on the first & last element, as simplified in 63ea7c41f06ac72fdf07894531487cfaae8eea85), multiline titles (or titles with mixed styling, e.g. h3 + body) actually don't have the correct margins. I believe multiline titles are an edge-case, since there is also the description, so the tradeoff might be alright.

When only the title (can be multiline!) is present, the title & CTA elements should be vertically centered in _View_ mode:
image

Alignment of title & image (before the fix):
image


packages/template-block/src/components/TemplateText.tsx
```diff
@@ -93,17 +94,30 @@ export const TemplateText = ({
         [appBridge, blockId, description, isEditing, saveDescription, templateTextKey],
     );
 
+    const titleClasses = useMemo(() => {
```
ragi96, line 97: Wouldn't use `useMemo` as its not a heavy operation nor does the `prop` often change

packages/template-block/src/components/TemplateText.tsx
```diff
@@ -93,17 +94,30 @@ export const TemplateText = ({
...
+        const removeTopMargin =
+            '[&>div>*:first-child]:!tw-mt-0 [&>div>[data-slate-editor="true"]>*:first-child]:!tw-mt-0';
+        const removeBottomMargin =
+            '[&>div>*:last-child]:!tw-mb-0 [&>div>[data-slate-editor="true"]>*:last-child]:!tw-mb-0';
```
ragi96, line 101: You can't rely on that, do something like here: https://github.com/Frontify/guideline-blocks/blob/59c20b9a0ba18a17d08198892de064bb8bb792b1/packages/callout-block/src/CalloutBlock.tsx#L85

Where we overwrite the theme settings, this you can overwrite the margin-bottom/top of all available text styles, and you don't have to overwrite them via tailwind

ragi96 (changes requested):

ryancarville (approved):
Other than @ragi96 comments, LGTM 👍

packages/template-block/src/components/TemplateText.tsx
findmind: Thanks for the suggestion! 💪

ragi96 (approved):
LGTM

PR #728, opened 2024-01-29
fix: Template Block, various improvements

findmind:
This PR adds following improvements to the template block:

- [x] Add CTA on preview click in view mode
- [x] Disable border by default -> up for debate, check with Guidelines team on what's preferred. Fixed for migration in https://github.com/Frontify/app-server/pull/18102
- [x] Improve padding-left in non-preview mode -> Not really possible, since this is dependent on the height of the text vs. the height of the button. If the text is a one-liner, adding padding is fine; if it's multiline it looks weird.
- [x] Decrease flex gap
- [x] Fix margin-collapse in Edit mode between title & description when `pageCount` is disabled
- [x] Fix low-hanging a11y issues
- [x] Fixed issues with button placement (button would not stick to the right in non-preview-mode, depending on previous settings)

Known issues:
- [ ] When `templateInput` has been used (aka a template has been chosen in the sidebar settings rather than in the block itself), subscriptions are dangling - this leads to duplicate requests, even when using the button within the block itself. Has to be fixed within `web-app`.


packages/template-block/src/components/PreviewImage.tsx
```diff
@@ -93,7 +93,7 @@ export const PreviewImage = ({
                         ? blockAssets.previewCustom[0].height
                         : 'auto'
                 }
-                alt={blockSettings.altText || undefined}
+                alt={blockSettings.altText || 'Template preview'}
```
ryancarville, line 96: `nit` nullish coalescing operator
```suggestion
                alt={blockSettings.altText ?? 'Template preview'}
```

ryancarville (approved):
Test and reviewed.  LGTM 👍 Just one `nit` about using `nullish coalescing operator` instead of a logical `OR`.

packages/template-block/src/hooks/useTemplateBlockData.ts
```diff
@@ -158,11 +158,16 @@ export const useTemplateBlockData = (appBridge: BlockProps['appBridge']) => {
...
+const getContentClasses = (
+    hasPreview: boolean,
+    textPositioning: TextPositioningType,
+    textAnchoringVertical: AnchoringType
+): string => {
+    const alignContentItems =
+        hasPreview && [TextPositioningType.Right, TextPositioningType.Left].includes(textPositioning)
+            ? verticalAlignmentToItemAlign[textAnchoringVertical]
+            : '';
+    return `tw-flex tw-gap-x-6 tw-gap-y-4 ${textPositioningToContentFlexDirection[textPositioning]} ${alignContentItems}`;
```
findmind, line 170: Here, I just added the `hasPreview` parameter, and added that to the `alignContentItems` condition (it's a bit hard to see). Also, I change `tw-gap-x-8` to `tw-gap-x-6`.

packages/template-block/src/components/TemplatePreview.tsx
```diff
@@ -60,7 +63,10 @@ export const TemplatePreview = ({
                         borderRadius,
                         border,
                         height,
+                        cursor: enableCta ? 'pointer' : 'default',
                     }}
```
ragi96, line 67: Maybe better to add `disabled` to the button, or is this not suitable, then you also don't need to handle the onClick...

findmind: Good catch; that's actually a leftover of when I was trying to use a div - which of course doesn't make semantical sense. It's fixed and simplified in 4c6ce2602243c786d0ed66d081f0040cb43c0aa9

ragi96 (approved):

PR #736, opened 2024-02-05
Template Block: Hide page count by default

findmind:
Since most templates use 1 page only, and when first creating block it says "0 pages", we'll disable the page-count by default.


ryancarville (approved):
LGTM 👍

ragi96 (approved):
LGTM

PR #737, opened 2024-02-05
feat(TemplateBlock): Use `getCurrentLoggedUser` to check if user is logged in [CU-8693qphk1]

findmind:
Fixes https://app.clickup.com/t/8693qphk1

👁️  Template Block should be hidden for users that are not logged in, because they can't create publications anyway.

⚠️  Notice regarding tests: I decided to mock a user with ID `0`, which evaluates to `false` in my code. I know this is not ideal, but at least we can test the actual behavior this way. As soon as a better solution is in place we should be able to easily switch it out.

🧪  How to test:
- Create a guideline with this version of template block (local block development)
- In the powerbar, go to the "Share" menu and get the "Public Link" (e.g. `https://dev.frontify.test/d/G9HPcwB3gvuk`)
- Open the public link e.g. in an Inkognito tab (w/o authenticated user) - the guidelines should be shown, but not the template block

image

Issue:
- [ ] I get a JSON.parse error from the HTTP library. In production mode, however, it works (aka the block is actually hidden). Not very elegant, though. -> Could be band-aided a bit by https://github.com/Frontify/web-app/pull/2791, but it's under consideration by Rafi & Alev (Shoutout to you! 🙏 )

### DIFF
--- a/packages/template-block/src/TemplateBlock.spec.ct.tsx
+++ b/packages/template-block/src/TemplateBlock.spec.ct.tsx
@@ -79,6 +79,20 @@ describe('Template Block', () => {
         cy.get(CTA_BUTTON_SELECTOR).should('exist');
     });
 
+    it('should not render a block if use is not authenticated', () => {
+        const [TemplateBlockWithStubs] = withAppBridgeBlockStubs(TemplateBlock, {
+            blockTemplates: {
+                template: [getTemplateDummyWithPages()],
+            },
+            user: undefined,
+        });
+
+        mount(<TemplateBlockWithStubs />);
+        cy.get(BLOCK_CONTAINER_SELECTOR).should('exist');
+        cy.wait(100);
+        cy.get(CARD_SELECTOR).should('not.exist');
+    });
+
     it('should render a block if template provided and block is in edit mode', () => {
         const [TemplateBlockWithStubs] = withAppBridgeBlockStubs(TemplateBlock, {
             editorState: true,
--- a/packages/template-block/src/TemplateBlock.spec.ct.tsx
+++ b/packages/template-block/src/TemplateBlock.spec.ct.tsx
@@ -1,7 +1,7 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
 import { mount } from 'cypress/react18';
-import { AssetDummy, TemplateDummy, withAppBridgeBlockStubs } from '@frontify/app-bridge';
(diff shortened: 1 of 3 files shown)


packages/template-block/src/hooks/useTemplateBlockData.ts
```diff
@@ -123,6 +124,12 @@ export const useTemplateBlockData = (appBridge: BlockProps['appBridge']) => {
     const hasPreview = preview !== PreviewType.None;
     const hasTitleOnly = !hasPreview && !hasPageCount && !hasRichTextValue(description);
 
+    useEffect(() => {
+        appBridge.getCurrentLoggedUser().then((user) => {
```
ragi96, line 128: Prefer await => will be foreced as soon we bump eslint-config

packages/template-block/src/TemplateBlock.spec.ct.tsx
```diff
@@ -79,6 +79,20 @@ describe('Template Block', () => {
...
+        const [TemplateBlockWithStubs] = withAppBridgeBlockStubs(TemplateBlock, {
+            blockTemplates: {
+                template: [getTemplateDummyWithPages()],
+            },
+            user: UserDummy.with(0),
+        });
+
+        mount(<TemplateBlockWithStubs />);
+        cy.get(BLOCK_CONTAINER_SELECTOR).should('exist');
+        cy.wait(100);
```
findmind, line 92: This wait is there to finish the React rendering first. I noticed that `.should('exist')` AND `.should('not.exist')` both were true - because Cypress was quicker than the React rendering, so first the div was not there, then it was. 😅

ragi96: Maybe we should migrate these tests to `vitest` (What we gonna do for new blocks anyway...)

ragi96 (approved):
LGTM, imo no need to adjust the web-app, so the other PR can be closed.

PS: i did a private release on product, to test it more properly

ryancarville (approved):
Code LGTM 👍 

I did not test personally, since @ragi96 did this thoroughly already.  🙏

packages/template-block/src/TemplateBlock.spec.ct.tsx
findmind: Definitely! I wasn't aware of the state here.. As it's only one file anyway I assume it's not gonna be extremely difficult to do that. Let's do that as a future enhancement.

PR #738, opened 2024-02-06
feat: image block updates

SamCreasey:
CU-8693kyqyc

Initial design changes for the image block:
- Moved "alt text" from the flyout menu into a separate component
- Added delete option but kept existing menu item


ragi96:
merged so somebody gets assigned, sorry didn't had it in my mind

utmishra (changes requested):
Hey @SamCreasey 
Visually, the changes look good 👍🏽 
There seems to be just one issue. Once I've set the alt text, I cannot remove it. If I delete the existing entry and save the alt text with blank text, I see that an empty `settings` payload is passed, thus the alt text is not updated.

packages/image-block/src/ImageBlock.tsx
```diff
@@ -44,7 +44,6 @@ export const ImageBlock = withAttachmentsProvider(({ appBridge }: BlockProps) =>
     const { openAssetChooser, closeAssetChooser } = useAssetChooser(appBridge);
     const isEditing = useEditorState(appBridge);
     const blockId = appBridge.getBlockId().toString();
```
ragi96, line 46:
```suggestion
    const blockId = String(appBridge.context('blockId').get());
```

packages/shared/src/components/EditAltTextFlyout/EditAltTextToolbarButton.tsx
```diff
@@ -0,0 +1,42 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
```
ragi96, line 1: Would be cool to have this one tested

SamCreasey: fixed

ragi96 (approved):
Next time please in `vitest`, as we will migrate them anyway

utmishra (approved):
Looks good 👍🏽

PR #742, opened 2024-02-12
fix: remove border radius when border is switched off

SamCreasey:
Removes the border radius setting from the sidebar when the border is switched off, as well as removing border radius styles in this same instance


fulopdaniel:
Hmmmm, the code looks good but are we sure that this is the expected behavior? Corner radiuses can be present without having borders, it feels a bit weird to make it depend on the border.

SamCreasey:
> Hmmmm, the code looks good but are we sure that this is the expected behavior? Corner radiuses can be present without having borders, it feels a bit weird to make it depend on the border.

This fixes an issue identified here by Martin https://app.clickup.com/t/86933q1m6 It appears that corder radius should not be visible when there is no border

fulopdaniel (approved):

PR #749, opened 2024-02-19
chore: update @frontify/guideline-blocks-settings@0.33.1

SamCreasey:
Update guideline-blocks settings to 0.33.1 and remove all instances of the `toolbarFlyoutItems` prop. There are no functional changes. For testing all blocks that had a block wrapper should behave the same as before.


ragi96 (approved):
LGTM

PR #750, opened 2024-02-19
fix(ImageBlock): gif loop count is not respected

fulopdaniel:
You can change a GIF's loop count using several tools, like https://ezgif.com/loop-count .

During our processing this is lost, and GIFs served via the frontify CDN loop infinitely. 
The image at the originURL still contains this property.

https://app.clickup.com/t/8693vp7d0


SamCreasey (approved):
Happy to have this as an interim solution, but would be great to see this addressed by the image worker team

PR #753, opened 2024-02-20
feat: lazyImages component

ragi96:


ragi96:
Close as its running into the same issue as with just `loading="lazy"`, as the IntersectionOberserver needs to be available before we load the images

PR #754, opened 2024-02-21
feat(ImageBlock): add asset viewer security settings

fulopdaniel:
https://app.clickup.com/t/8693cyyce

### DIFF
--- a/packages/image-block/src/ImageBlock.spec.ct.tsx
+++ b/packages/image-block/src/ImageBlock.spec.ct.tsx
@@ -16,6 +16,7 @@ import { Security } from '@frontify/guideline-blocks-settings';
 
 const ImageBlockSelector = '[data-test-id="image-block"]';
 const ImageBlockImageWrapperSelector = '[data-test-id="image-block-img-wrapper"]';
+const ImageBlockAssetViewerButtonSelector = '[data-test-id="image-block-asset-viewer-button"]';
 const ImageBlockImageSelector = '[data-test-id="image-block-img"]';
 const ImageBlockCaption = '[data-test-id="image-caption"]';
 const PlaceholderSelector = '[data-test-id="block-inject-button"]';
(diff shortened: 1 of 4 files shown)


packages/image-block/src/ImageBlock.spec.ct.tsx
```diff
@@ -81,6 +82,84 @@ describe('Image Block', () => {
...
+            privacySettings: {
+                assetViewerEnabled: false,
+                assetDownloadEnabled: true,
+            },
+            blockAssets: {
+                [IMAGE_ID]: [AssetDummy.with(1)],
+            },
+        });
+        mount(<ImageBlockWithStubs />);
+        cy.get(ImageBlockImageSelector).should('exist');
```
ragi96, line 101: Usesless to check if the ImageBlockImageSelector exists same for the other tests

packages/image-block/src/components/Image.tsx
```diff
@@ -24,8 +25,12 @@ export const ImageComponent = ({ image, blockSettings, isEditing, appBridge }: I
     const { open } = useAssetViewer(appBridge);
     const link = blockSettings?.hasLink && blockSettings?.linkObject?.link && blockSettings?.linkObject;
     const imageStyle = getImageStyle(blockSettings, image.width);
+    const { assetViewerEnabled: globalAssetViewerEnabled } = usePrivacySettings(appBridge);
```
ragi96, line 28: This hook is now used twice, can you use it just once and pass the assetViewEnabled just as a prop?

packages/image-block/src/components/Image.tsx
```diff
@@ -49,23 +54,33 @@ export const ImageComponent = ({ image, blockSettings, isEditing, appBridge }: I
...
+            return (
+                <button data-test-id="image-block-asset-viewer-button" {...props} onClick={() => open(image)}>
+                    {Image}
+                </button>
+            );
+        }
+        return Image;
+    };
+
+    return getImageComponent();
```
ragi96, line 83:
```suggestion
        if (isEditing) {
        return Image;
    }

    if (link) {
        return (
            <a
                {...props}
                href={link.link.link}
                target={link.openInNewTab ? '_blank' : undefined}
                rel={link.openInNewTab ? 'noopener noreferrer' : 'noreferrer'}
            >
                {Image}
            </a>
        );
    }

    if (isAssetViewerEnabled) {
        return (
            <button data-test-id="image-block-asset-viewer-button" {...props} onClick={() => open(image)}>
                {Image}
            </button>
        );
    }

    return Image;
```

Still not good, but better

packages/image-block/src/ImageBlock.spec.ct.tsx
fulopdaniel: Not useless:

1. Without it the "not.exists" tests are false positive, because cypress runs the tests too fast, and asserts the not.exists without even waiting for the dom to be properly rendered.
2. It makes sure that the img tag itself is rendered.

packages/image-block/src/components/Image.tsx
fulopdaniel: Nice, adjusted

packages/image-block/src/components/Image.tsx
fulopdaniel: Adjusted

packages/image-block/src/ImageBlock.spec.ct.tsx
ragi96: Ah I see, then its fine :D

ragi96 (approved):

PR #755, opened 2024-02-22
fix(Image-block): apply title when rte is empty

SamCreasey:
When an image is first uploaded the name and alt-text will be set to the uploaded image title, and to the filename if the title does not exist. When a second image is uploaded neither of these are updated

### DIFF
--- a/packages/image-block/src/ImageBlock.tsx
+++ b/packages/image-block/src/ImageBlock.tsx
@@ -26,6 +26,7 @@ import {
     IconTrashBin20,
     LoadingCircle,
     MenuItemStyle,
+    generateRandomId,
 } from '@frontify/fondue';
 import { useEffect, useState } from 'react';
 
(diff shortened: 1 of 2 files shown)


fulopdaniel:
I would go with a bit different approach.

I would pass `image?.fileName` as a prop to the `ImageCaption` component.

```
<ImageCaption defaultName={image?.fileName ?? ""} {...otherProps} />
```

And then in ImageCaption do this:

```
<RichTextEditor
    key={defaultName}
    value={name ?? convertToRteValue(TextStyles.imageTitle, defaultName, 'center')}
    {...otherProps}
/>
```

The upside with this approach is
1, you don't need to await the blocksettings change, which I feel is a bit not ideal already.
2, you can change images and the filename will change accordingly in the title up until the point you decide to manually edit the title.

What do you think?

SamCreasey:
> I would go with a bit different approach.
> 
> I would pass `image?.fileName` as a prop to the `ImageCaption` component.
> 
> ```
> <ImageCaption defaultName={image?.fileName ?? ""} {...otherProps} />
> ```
> 
> And then in ImageCaption do this:
> 
> ```
> <RichTextEditor
>     key={defaultName}
>     value={name ?? convertToRteValue(TextStyles.imageTitle, defaultName, 'center')}
>     {...otherProps}
> />
> ```
> 
> The upside with this approach is 1, you don't need to await the blocksettings change, which I feel is a bit not ideal already. 2, you can change images and the filename will change accordingly in the title up until the point you decide to manually edit the title.
> 
> What do you think?

I like this idea, especially part 2. The downside of this approach though is that the name RTE will always unmount when the image changes even if the text content is correct. I also am not sure if this would be expected behaviour so I will check with design first

SamCreasey:
Design feedback: "Let’s use the same pattern as we have it in the audio or image grid block. The title & alt tag should not be changed, when an asset is replaced." 
Pr has been updated according to this behaviour description

fulopdaniel (approved):

PR #760, opened 2024-02-27
fix(ImageBlock): update background and resolution behavior

fulopdaniel:
Background now applies to the block.

Image resolution request is as follows:

Request the 800px or 1600px width version of the image depending on screen resolution, except if the original image width is not wide enough for that. In that case request the original image width.

https://github.com/Frontify/guideline-blocks/pull/760#issue-2156803949


packages/image-block/src/components/helpers.ts
```diff
@@ -90,7 +90,7 @@ export const getTotalImagePadding = (blockSettings: Settings): CSSProperties =>
 const getPadding = (blockSettings: Settings): string =>
     blockSettings.hasCustomPadding ? blockSettings.paddingCustom : paddingValues[blockSettings.paddingChoice];
 
-export const getImageStyle = (blockSettings: Settings, width: string): CSSProperties => {
+export const getImageWrapperStyle = (blockSettings: Settings): CSSProperties => {
```
SamCreasey, line 93: https://www.loom.com/share/7c292bd3939a4d278eeb1147479cdc46 The background appears to be visible on viewmode underneath long images

packages/image-block/src/components/Image.tsx
```diff
@@ -47,7 +49,7 @@ export const ImageComponent = ({
             loading="lazy"
             src={src}
             alt={blockSettings.altText || undefined}
-            style={imageStyle}
+            style={{ maxWidth: image.width }}
```
SamCreasey, line 52: After looking into this a bit I wonder if it would be a good idea to add 'height' and 'width' attributes to the image. They seem to only be used to maintain the aspect ratio but it helps reduce layout shifting

packages/image-block/src/components/Image.tsx
```diff
@@ -30,15 +30,17 @@ export const ImageComponent = ({
 }: ImageProps & { isAssetViewerEnabled: boolean }) => {
     const { open } = useAssetViewer(appBridge);
     const link = blockSettings?.hasLink && blockSettings?.linkObject?.link && blockSettings?.linkObject;
-    const imageStyle = getImageStyle(blockSettings, image.width);
     const { isFocused, focusProps } = useFocusRing();
 
+    const devicePixelRatio = Math.max(1, window?.devicePixelRatio ?? 1);
```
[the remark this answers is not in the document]
fulopdaniel: Actually there were some issues reported a while ago that images were requested with a devicePixelRatio of around 0.9. It was super weird and this was implemented as a safety net to always request with at least a DPR of 1.

packages/image-block/src/components/helpers.ts
fulopdaniel: Thanks, fixed by making the container flex.

packages/image-block/src/components/Image.tsx
fulopdaniel: I added the height and width attributes, I like the idea. My only concern is that it might need some refactoring once the page container is not fixed to 800px but for now it seems to be an improvement.

SamCreasey: I actually meant directly setting them on the img element, so it shouldn't cause us any trouble

```
<img width={image.width} height={image.height}/>
```

fulopdaniel: sure, changed to that

SamCreasey (approved):

PR #761, opened 2024-02-27
feat: Template block button styling [CU-8693y5d2j]

findmind:
This PR adds support to choose the CTA button styling between Primary, Secondary or Tertiary; before it was hardcoded to primary. Closes https://app.clickup.com/t/8693y5d2j

- [x] Fix styling issues #762 
- [x] Add test

image

### DIFF
--- a/packages/template-block/src/TemplateBlock.spec.ct.tsx
+++ b/packages/template-block/src/TemplateBlock.spec.ct.tsx
@@ -56,6 +56,15 @@ const getTemplateDummyWithPages = () => {
     return templateDummy;
 };
 
+const getValueFromCssVariable = (cssVarName: string, styleName: string) => {
+    const dummy = document.createElement('span');
+    dummy.style.setProperty(styleName, `var(${cssVarName})`);
+    document.body.appendChild(dummy);
+    const value = window.getComputedStyle(dummy).getPropertyValue(styleName).trim();
+    dummy.remove();
+    return value;
+};
+
 describe('Template Block', () => {
     it('should render an empty slate if no template provided', () => {
         const [TemplateBlockWithStubs] = withAppBridgeBlockStubs(TemplateBlock);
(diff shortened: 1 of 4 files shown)


packages/template-block/src/TemplateBlock.spec.ct.tsx
```diff
@@ -56,6 +56,15 @@ const getTemplateDummyWithPages = () => {
...
 };
 
+const getValueFromCssVariable = (cssVarName: string, styleName: string) => {
+    const dummy = document.createElement('span');
+    dummy.style.setProperty(styleName, `var(${cssVarName})`);
+    document.body.appendChild(dummy);
+    const value = window.getComputedStyle(dummy).getPropertyValue(styleName).trim();
+    dummy.remove();
+    return value;
+};
```
findmind, line 66: Yes, this is a bit wild, and the values for secondary and primary in this case are actually the same. However, imho it's actually testing the applied values.. Or would you sth like `data-text-button-style="secondary"` to the DOM element that can be tested?

ragi96: Would just test if the button styles are applied.

So if you wanna test `font-size` you just check that the value is `var(--f-theme-settings-button-primary-font-size)` and not the value of the var

findmind: Doesn't work, it only gets the actual value rather than the variable itself. I tried to search how to get the variable as a string, but didn't find anything. So I'm applying the same variable and comparing the outcome basically.

image

findmind: Damn that's genious, can't believe I haven't found this approach 🙈 Keep on learning 🚀 

Fixed in #[49fc2ed](https://github.com/Frontify/guideline-blocks/pull/761/commits/49fc2ed2bc9dc54406c68ed8fda6d1e4c7f0dd06)

mikeyrayvon (approved):
g2g from my side

SamCreasey (commented):
Works as expected, nice! 👍

SamCreasey (approved):

PR #763, opened 2024-02-28
feat(UIPatternBlock): add css preprocessor support for vanilla

fulopdaniel:


packages/ui-pattern-block/src/helpers/files.ts
```diff
@@ -1,9 +1,21 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
 import type { SandpackFiles } from '@codesandbox/sandpack-react/types';
-import { SandpackTemplate } from '../types';
+import { Preprocessor, SandpackTemplate } from '../types';
+
+const getCssExtension = (preprocessor: Preprocessor) => {
```
ragi96, line 6: Please add a unit test for this one

ragi96 (commented):
For the migration this is good enough 👍 

But I think in the long run we should drop all that defaultFileHandling etc from the block and just implement https://sandpack.codesandbox.io/docs/advanced-usage/components#file-explorer

So we don't need to care about this at all

ragi96 (approved):

PR #764, opened 2024-02-29
feat(ImageBlock): remove 800px hardcoding

fulopdaniel:
This approach always requests images as wide as the container. This means smaller images for mobiles, and better support for the upcoming themes where container width might not always be 800 px.

### DIFF
--- a/packages/image-block/src/components/Image.tsx
+++ b/packages/image-block/src/components/Image.tsx
@@ -13,6 +13,7 @@ import { useFocusRing } from '@react-aria/focus';
 import { AppBridgeBlock, Asset, useAssetViewer, usePrivacySettings } from '@frontify/app-bridge';
 import { getImageWrapperStyle, getTotalImagePadding } from './helpers';
 import { FOCUS_STYLE } from '@frontify/fondue';
+import { useState } from 'react';
 
 type ImageProps = {
     image: Asset;
@@ -27,14 +28,14 @@ export const ImageComponent = ({
     isEditing,
     appBridge,
     isAssetViewerEnabled,
-}: ImageProps & { isAssetViewerEnabled: boolean }) => {
+    containerWidth,
+}: ImageProps & { isAssetViewerEnabled: boolean; containerWidth: number }) => {
     const { open } = useAssetViewer(appBridge);
     const link = blockSettings?.hasLink && blockSettings?.linkObject?.link && blockSettings?.linkObject;
     const { isFocused, focusProps } = useFocusRing();
 
     const devicePixelRatio = Math.max(1, window?.devicePixelRatio ?? 1);
-    const imageWidthToRequest = Math.min(800 * devicePixelRatio, image.width);
-
+    const imageWidthToRequest = Math.min(containerWidth * devicePixelRatio, image.width);
     // Gif images can have a loop count property
     // Which is lost during our image processing
     const src =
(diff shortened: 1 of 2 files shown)


ragi96 (commented):
Wouldn't it be better to just introduce 

`sizes` and `srcset`, so we have it perfect?

fulopdaniel:
> Wouldn't it be better to just introduce
> 
> `sizes` and `srcset`, so we have it much better?

We can look into it but I think the biggest issue would be that `sizes` support media queries that rely on screen width but we need the container width, which depends on the theme..

Also this way it is a bit more precise, as it exactly requests the size needed.

packages/image-block/src/components/Image.tsx
```diff
@@ -102,16 +104,20 @@ export const Image = ({ image, appBridge, blockSettings, isEditing }: ImageProps
         <div
             style={imageWrapperStyle}
             data-test-id="image-block-img-wrapper"
+            ref={(el) => setContainerWidth(el?.offsetWidth)}
```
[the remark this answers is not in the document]
fulopdaniel: Added the test 👍

fulopdaniel:
> > tainer width, which depends on the theme..
> 
> Yeah we would need to calculate it perfectly depending on the container size and transform it into whats needed on viewport size 🤯

Not sure if we can do that. I mean let's say currently the container width is 800px and screen width is 1300px. How do you calculate what screen width will result in a container width of 600px. (in a universal, theme-agnostic way)

ragi96:
> > > tainer width, which depends on the theme..
> > 
> > 
> > Yeah we would need to calculate it perfectly depending on the container size and transform it into whats needed on viewport size 🤯
> 
> Not sure if we can do that. I mean let's say currently the container width is 800px and screen width is 1300px. How do you calculate what screen width will result in a container width of 600px. (in a universal, theme-agnostic way)

Yeah we sure need to check about the possibility, its not easy. 
Otherway would be to wait for that: https://github.com/w3c/csswg-drafts/issues/5889
Give it 2 or 3 more years

packages/image-block/src/components/Image.tsx
```diff
@@ -27,14 +28,14 @@ export const ImageComponent = ({
...
+    containerWidth,
+}: ImageProps & { isAssetViewerEnabled: boolean; containerWidth: number }) => {
     const { open } = useAssetViewer(appBridge);
     const link = blockSettings?.hasLink && blockSettings?.linkObject?.link && blockSettings?.linkObject;
     const { isFocused, focusProps } = useFocusRing();
 
     const devicePixelRatio = Math.max(1, window?.devicePixelRatio ?? 1);
-    const imageWidthToRequest = Math.min(800 * devicePixelRatio, image.width);
-
+    const imageWidthToRequest = Math.min(containerWidth * devicePixelRatio, image.width);
```
SamCreasey, line 38: Just out of curiosity do you know if there would be a benefit from having a list of image sizes, and then picking the lowest size that fits the container? I'm wondering if this would help somehow with caching the images. For example:
We could have sizes [240, 360, 480, 720 , 1080, 1440] and. if the container is 800px it will request the 1080 version. Not sure if this would help

SamCreasey: The reason why I think this might be something good to add is also because at the minute if the block changes size the containerWidth state does not update. Especially when it comes to sections I can imagine a situation where the user has an image block as one of 3 blocks on a row. Since the user could load the page at a certain width, resizing the page could cause the image block to wrap and then actually expand from it's original size, since it will now be on a row by itself and the image will be too low quality.

We could run a resizeObserver to see which size in the array matches best. Since it would be using the array value it would have better performance than using the raw width as it would only trigger a state change when a new array size is reached (eg: container.width: 300 => 400 changes the image width from '360' to '480', and only modifies the state once. 300 => 350 causes no state changes to occur). 
Minor edge case though so it could be a separate pr.

fulopdaniel: I like this approach, seems like a sweet spot between using `srcset` and hardcoding one value. Small concern of mine is how much would it hurt the performance to attach a resize observer for each image block, but will need to discover this.

For this I would prefer a separate PR and ticket.

SamCreasey: Sure, sounds good 👍

SamCreasey (approved):

PR #770, opened 2024-03-05
feat(ImageBlock): make images responsive

fulopdaniel:
If accepted, `ResponsiveImage` and `useImageContainer` could potentially move to brand-sdk.


packages/image-block/src/hooks/useImageContainer.tsx
```diff
@@ -0,0 +1,39 @@
...
+            debounce((entries) => {
+                const container = entries[0];
+                const newContainerWidth = container.contentRect.width;
+                const oldContainerWidth = containerWidth ?? 0;
+                const containerWidthHasGrown = oldContainerWidth < newContainerWidth;
+                if (containerWidthHasGrown) {
+                    setContainerWidth(container.contentRect.width);
+                }
+            }, 100)
+        );
```
SamCreasey, line 25: My only concern with this is that it is a lot harder to cache images. I think it makes more sense to use the pre-determined array of sizes approach, although I'm unaware if there are any rendering issues if for example you render a 1080px image in an 800px container.

SamCreasey: This would also mean changing the name of the hook probably

packages/image-block/src/hooks/useImageContainer.tsx
```diff
@@ -0,0 +1,39 @@
...
+        }
+
+        const containerObserver = new ResizeObserver(
+            debounce((entries) => {
+                const container = entries[0];
+                const newContainerWidth = container.contentRect.width;
+                const oldContainerWidth = containerWidth ?? 0;
+                const containerWidthHasGrown = oldContainerWidth < newContainerWidth;
+                if (containerWidthHasGrown) {
+                    setContainerWidth(container.contentRect.width);
```
SamCreasey, line 22:
```suggestion
                    setContainerWidth(newContainerWidth);
```

packages/image-block/src/components/ResponsiveImage.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+type ResponsiveImageProps = {
+    image: Asset;
+    containerWidth: number;
+    altText?: string;
+    allowWebp?: boolean;
+    quality?: number;
+};
+
+export const ResponsiveImage = ({
```
SamCreasey, line 13: If we can I would add tests now and then it is tested and easily portable to guideline-blocks-settings

fulopdaniel: Yeah, added a bunch of tests 👍

packages/image-block/src/hooks/useImageContainer.tsx
fulopdaniel: Adjusted

packages/image-block/src/hooks/useImageContainer.tsx
fulopdaniel: Yeah that makes sense. To make it flexible, I made it that the container width is always rounded to the next hundred. So possible values are 100, 200, 300, 400, etc...

What do you think?

packages/image-block/src/components/ResponsiveImage.spec.ct.tsx
```diff
@@ -0,0 +1,108 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
```
ragi96, line 1: Would be nice to have these new tests written with vitest

fulopdaniel: Added the tests in vitest

packages/image-block/src/hooks/useImageContainer.tsx
SamCreasey: Honestly I think this will come from trial and error to find a nice in-between with image size and number of requests, but for now it works really well

packages/image-block/src/components/ResponsiveImage.spec.tsx
```diff
@@ -0,0 +1,112 @@
...
+        );
+    });
+
+    it('should use original image width and height as attributes', () => {
+        const { getByTestId } = render(<ResponsiveImage image={HIGH_RES_ASSET} containerWidth={800} />);
+        expect(getByTestId(ResponsiveImageSelector).getAttribute('width')).toBe('2000');
+        expect(getByTestId(ResponsiveImageSelector).getAttribute('height')).toBe('1000');
+    });
+
+    it('should use original image width as max height', () => {
```
SamCreasey, line 96:
```suggestion
    it('should use original image width as max width', () => {
```

packages/image-block/src/hooks/useImageContainer.tsx
```diff
@@ -0,0 +1,40 @@
...
+    useEffect(() => {
+        if (!containerRef.current) {
+            return;
+        }
+
+        const containerObserver = new ResizeObserver(
+            debounce((entries) => {
+                const container = entries[0];
+                const newContainerWidth = roundToNextHundred(container.contentRect.width);
+                const oldContainerWidth = roundToNextHundred(containerWidth ?? 0);
```
SamCreasey, line 20: containerWidth is always a value of 100 so I think this is unneccessary to wrap in the fn here

packages/image-block/src/components/ResponsiveImage.spec.tsx
```diff
@@ -0,0 +1,112 @@
...
+};
+
+describe('ResponsiveImage', () => {
+    beforeAll(() => {
+        configure({ testIdAttribute: 'data-test-id' });
+    });
+
+    afterEach(() => {
+        cleanup();
+    });
```
SamCreasey, line 26: Can we move this to a vite setup file? It will be great to have for all blocks

SamCreasey (approved):
Awesome!

PR #771, opened 2024-03-06
feat: Template Block Performance Improvements [CU-86940ffk3]

findmind:
This PR reduces the amount of `/asset` API requests by only using the `useBlockAssets` hook once, instead of 3x. It also cleans up data handling, esp. regarding the `previewCustom`.

What to test: Set the preview to "Custom", and check that there actually is only 1 `/asset` API call. Before this PR, there should be 3 API calls.

Would be nice:
- [ ] Don't use `useBlockAssets` at all if preview is not set to custom -> blocked because `handleNewPublication` needs the custom `previewUrl`; could be circumvented by using the `useBlockAsset` hook conditionally only if the custom preview is actually set, but this gets extremely fiddly with state management.

### DIFF
--- a/packages/template-block/src/TemplateBlock.tsx
+++ b/packages/template-block/src/TemplateBlock.tsx
@@ -39,8 +39,8 @@ export const TemplateBlock = ({ appBridge }: BlockProps): ReactElement => {
     const handleNewPublication = async () => {
         if (selectedTemplate !== null) {
             const previewUrl =
-                preview === PreviewType.Custom && previewCustom?.length > 0
-                    ? previewCustom[0].previewUrl
+                preview === PreviewType.Custom && previewCustom
+                    ? previewCustom.previewUrl
                     : selectedTemplate.previewUrl;
 
             const options: OpenNewPublicationPayload = {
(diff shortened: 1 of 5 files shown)


SamCreasey (approved):
Awesome work. This is a great improvement. I am just wondering if it's also worth replacing the previewUrl {width} with the correct sizing. Our team is developing a ResponsiveImage component. I wonder if we could work together to use it here too, although this could be done in a separate pr

ragi96:
> Awesome work. This is a great improvement. I am just wondering if it's also worth replacing the previewUrl {width} with the correct sizing. Our team is developing a ResponsiveImage component. I wonder if we could work together to use it here too, although this could be done in a separate pr

I already told @findmind that they should reuse the responsiveImage component, as soon its finished (now).

ryancarville (approved):
Nice performance update! 
✅ Code looking good
✅ function test preforms as expected

findmind:
@SamCreasey @ragi96 thanks for the input! Sure, let's use that. The image logic is relatively complex, with all the positioning logic etc. that's there, but I'm sure it can be simplified, esp. if that stuff is already covered by your new component?! 

What honestly bugs me more is that we can't easily do the `/asset` call conditionally depending on the custom preview setting. I played around a bit with putting the `useBlockAssets` hook into a shadow component that I only render conditionally, but because the custom preview is also used in the `handleNewPublication` function it got really ugly and looks like a workaround from a mile.

Imho the performance issues with many blocks are anyways not only related to the template block, so batch-fetching the assets (I bounced a couple of ideas there already) and esp. getting rid of the `/user/info` call will hugely help! ⛑️

PR #772, opened 2024-03-07
fix(ImageBlock): alignment is not applied

fulopdaniel:


utmishra (approved):
Looks good, alignment gets applied as expected in view mode

PR #779, opened 2024-03-11
fix(TemplateBlock): Replace `appBridge.getCurrentLoggedUser()` with class-check in DOM [CU-86940ffk3]

findmind:
This improves performance as we can get rid of the `/user/info` call. 💪 🐎 🚀 

- [x] Adapt tests

⚠️ This is only a temporary workaround; will be removed once the appBridge implementation has improved.

Fixes https://app.clickup.com/t/86940ffk3


ragi96 (approved):
Approved as an intermediate solution

PR #780, opened 2024-03-11
fix(ImageBlock): default quality was increased unintentionally

fulopdaniel:
The default quality is 75 when omitted.


ragi96 (approved):
Fix the tests and approved :)

PR #781, opened 2024-03-11
feat(DoDontBlock): performance improvement

fulopdaniel:
1. Make one block assets request and pass down to items
2. Change `updateAssetIdsFromKey` to `addAssetIdsToKey` or `deleteAssetIdsFromKey`
3. Improve DnD Sort context: prevent thousands of rerenders when dragging is active. Every px change resulted in rerendering the whole tree.
4. Extend ResponsiveImage with classname and style to make it more customisable.
5. Patch pnpm dependency like in brand-sdk, otherwise vite tests don't run

Waiting for feedback on whether we should turn the image mode on in `settings.ts`


package.json
```diff
@@ -48,6 +48,9 @@
     "pnpm": {
         "overrides": {
             "@codesandbox/sandpack-react@^2.12.1>@codesandbox/sandpack-client": "2.10.0"
+        },
+        "patchedDependencies": {
```
SamCreasey, line 52: What was this for? (Just curious)

fulopdaniel: Long story, but vite tests are not running because of a dependency of @udecode/plate is missing import assertions. There have been attempts to solve this within plate (https://github.com/udecode/plate/pull/2991) but their build process removed the "assert" keyword that was added. For now until this is fixed we patched it internally both in in brand-sdk and in web-app.

SamCreasey:
Screenshot 2024-03-12 at 18 04 56
Each do and don't item updates if you type into any RTE. Wrapping the item in memo is not doing anything right now. We should make sure these props are stable

packages/dos-donts-block/src/components/ImageComponent.tsx
```diff
@@ -70,11 +75,10 @@ const ImageComponent = ({
                 </div>
             </div>
         ) : null;
-    };
 
     return (
         <div className="tw-mb-3">
-            {(!src && isEditing) || isUploadLoading ? (
+            {(!image && isEditing) || isUploadLoading ? (
```
SamCreasey, line 81: Unrelated to your changes but is this really the expected behaviour? That the inject button appears in view mode conditionally when uploading?

packages/shared/src/components/ResponsiveImage/ResponsiveImage.tsx
```diff
@@ -39,13 +44,13 @@ export const ResponsiveImage = ({
...
-            className="tw-flex tw-w-full"
+            className={joinClassNames(['tw-flex tw-w-full', className])}
             loading="lazy"
             src={sourceOptimised}
             width={imageWidth}
             height={imageHeight}
-            style={{ maxWidth: imageWidth }}
-            alt={altText}
+            style={style}
+            alt={altText ?? ''}
```
SamCreasey, line 53: Can be done in a different pr but I think it would actually be better to enforce altText as a required string type. A user could add an empty string but this would force people to at least think about it, as it's very commonly a forgotten attribute

packages/image-block/src/components/Image.tsx
```diff
@@ -34,7 +34,14 @@ export const ImageComponent = ({
     const link = blockSettings?.hasLink && blockSettings?.linkObject?.link && blockSettings?.linkObject;
     const { isFocused, focusProps } = useFocusRing();
 
-    const Image = <ResponsiveImage image={image} containerWidth={containerWidth} altText={blockSettings.altText} />;
+    const Image = (
+        <ResponsiveImage
+            image={image}
+            containerWidth={containerWidth}
+            altText={blockSettings.altText}
+            style={{ maxWidth: image.width }}
```
SamCreasey, line 42: I don't understand why this was required as a change? It seems like it should have been working before and this behaviour is pretty important for the ResponsiveImage component

packages/shared/src/components/ResponsiveImage/ResponsiveImage.tsx
```diff
@@ -1,15 +1,18 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
-import React from 'react';
+import React, { CSSProperties } from 'react';
```
SamCreasey, line 3:
```suggestion
import React, { type CSSProperties } from 'react';
```

packages/dos-donts-block/src/components/ImageComponent.tsx
```diff
@@ -37,30 +37,35 @@ const ImageComponent = ({
...
+                ref={setContainerRef}
+                className={joinClassNames([
+                    'tw-relative tw-w-full',
+                    isEditing && (isDragging ? 'tw-cursor-grabbing' : 'tw-cursor-grab'),
+                ])}
+            >
                 <div
                     style={{ height: imageHeight, borderRadius, border, background }}
                     className="tw-w-full tw-overflow-hidden"
                 >
```
SamCreasey, line 53: Can this be merged into the parent div?

packages/dos-donts-block/src/DoDontItem.tsx
```diff
@@ -20,322 +20,308 @@ import {
...
-                                ],
-                                [
-                                    {
-                                        title: 'Delete',
-                                        icon: <IconTrashBin20 />,
-                                        onClick: onRemoveSelf,
-                                    },
-                                ],
+    return (
+        <div className="tw-bg-base tw-relative" style={{ ...(!isDragging ? { zIndex: undefined } : {}) }}>
```
SamCreasey, line 171: Is this style actually doing anything anymore?

packages/dos-donts-block/src/DoDontItem.tsx
```diff
@@ -344,25 +330,31 @@ export const SortableDoDontItem = (props: SortableDoDontItemProps) => {
...
-            key={id}
-            ref={setNodeRef}
-            {...props}
-            isDragging={isDragging}
-            replaceWithPlaceholder={isDragging}
-            transformStyle={transformStyle}
-            draggableProps={draggableProps}
-            setActivatorNodeRef={setActivatorNodeRef}
-        />
+        <div style={transformStyle} ref={setNodeRef}>
```
SamCreasey, line 349: Previously the transformStyle was being overwritten with zIndex:0, not sure if this is now not required here

packages/dos-donts-block/src/DoDontItem.tsx
```diff
@@ -20,322 +20,308 @@ import {
...
+                    />
+                )}
+                <div data-test-id="dos-donts-content" className={style === DoDontStyle.Icons ? 'tw-mt-3' : 'tw-mt-2'}>
+                    {memoizedRichTextEditor}
+                </div>
+            </BlockItemWrapper>
+            <div
+                className={joinClassNames([
+                    !replaceWithPlaceholder && 'tw-hidden',
+                    'tw-absolute tw-h-full tw-left-0 tw-top-0 tw-w-full tw-border-2 tw-border-box-selected-strong tw-border-dashed tw-rounded-[4px] tw-bg-box-selected-hover',
```
SamCreasey, line 319:
```suggestion
                    'tw-absolute tw-h-full tw-left-0 tw-top-0 tw-w-full tw-border-2 tw-border-box-selected-strong tw-border-dashed tw-rounded-sm tw-bg-box-selected-hover',
```

packages/dos-donts-block/src/DoDontItem.tsx
```diff
@@ -20,322 +20,308 @@ import {
...
-                                        } as CSSProperties
-                                    }
-                                    value={title}
-                                    disabled={!editing}
-                                    aria-label="Title"
-                                    placeholder={editing ? 'Add a title' : ''}
-                                    className="tw-text-s tw-pointer-ev tw-w-full tw-placeholder-[var(--placeholder-color)] placeholder:tw-opacity-30 tw-placeholder-opacity-30 tw-bg-transparent tw-resize-none tw-text-text-weak tw-break-words tw-outline-none tw-whitespace-pre-wrap"
-                                />
-                            </span>
+                    {style === DoDontStyle.Icons && (title || body || editing) && (
```
SamCreasey, line 249: body is always there I think because it's the RTE value

fulopdaniel: Initially it's an empty string which is falsy. That is why when you add an empty item the icons don't show up in view mode.

SamCreasey: Logic seems inconsistent though as if you type into the RTE and then backspace it will always appear after that

packages/image-block/src/components/Image.tsx
fulopdaniel: I figured this part is a bit opinionated, as in the Do & Don't block we don't want this behavior. If the object-fit is set to fill, then we want the image to fill the whole container regardless of the width of the original image.

So I introduced className and style prop to the ResponsiveImage to make it customisable to each use case.

packages/dos-donts-block/src/DosDontsBlock.tsx
```diff
@@ -37,13 +36,31 @@ import {
...
+    const [blockSettings] = useBlockSettings<Settings>(appBridge);
+
+    const shouldFetchAssets =
+        blockSettings.mode === BlockMode.TEXT_AND_IMAGE ||
+        blockSettings.hasCustomDoIcon ||
+        blockSettings.hasCustomDontIcon;
+
+    const Provider = shouldFetchAssets ? AssetsProvider : NoAssetsProvider;
+
+    return (
```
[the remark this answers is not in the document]
fulopdaniel: good one, implemented this way

packages/dos-donts-block/src/AssetsProvider.tsx
```diff
@@ -0,0 +1,30 @@
...
+    const { blockAssets, addAssetIdsToKey, deleteAssetIdsFromKey } = useBlockAssets(appBridge);
+    return (
+        <AssetsContext.Provider value={{ blockAssets, addAssetIdsToKey, deleteAssetIdsFromKey }}>
+            {children}
+        </AssetsContext.Provider>
+    );
+};
+
+export const NoAssetsProvider = ({ children }: { children: ReactNode }) => {
+    return <AssetsContext.Provider value={{}}>{children}</AssetsContext.Provider>;
```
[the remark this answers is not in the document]
fulopdaniel: Fixed

packages/dos-donts-block/src/DosDontsBlock.tsx
```diff
@@ -189,19 +206,23 @@ export const DosDontsBlock: FC<BlockProps> = ({ appBridge }) => {
...
+                setSelectedFiles(null);
+            });
+        }
     };
 
     const removeItemById = (itemId: string) => {
+        const itemToRemove = localItems.find((item) => item.id === itemId);
         const newItems: Item[] = localItems.filter((item) => item.id !== itemId);
+        if (itemToRemove?.imageId && deleteAssetIdsFromKey) {
+            deleteAssetIdsFromKey('itemImages', [itemToRemove?.imageId]);
```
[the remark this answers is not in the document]
fulopdaniel: Added

packages/dos-donts-block/src/components/ImageComponent.tsx
fulopdaniel: nice catch, updated it to always require `isEditing`

packages/shared/src/components/ResponsiveImage/ResponsiveImage.tsx
fulopdaniel: I included it in this pr 👍

packages/shared/src/components/ResponsiveImage/ResponsiveImage.tsx
fulopdaniel: fixed

packages/dos-donts-block/src/components/ImageComponent.tsx
fulopdaniel: Yep, merged

packages/dos-donts-block/src/components/ImageComponent.tsx
```diff
@@ -37,30 +37,35 @@ const ImageComponent = ({
...
-                        ])}
-                        src={src.replace('{width}', width.toString())}
-                        alt=""
-                    />
+                    {containerWidth && (
+                        <ResponsiveImage
+                            containerWidth={containerWidth}
+                            image={image}
+                            className={joinClassNames([
+                                'tw-h-full',
```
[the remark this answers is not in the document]
fulopdaniel: We could introduce this object-fit prop but I think it might not add too much, so for now I'd prefer to keep it simple with just the className and style tag being exposed.

packages/dos-donts-block/src/DoDontItem.tsx
fulopdaniel: Nope, removed

packages/dos-donts-block/src/DoDontItem.tsx
fulopdaniel: Fixed

packages/dos-donts-block/src/DoDontItem.tsx
fulopdaniel: Yep, changed the condition to `hasRichtextValue(body)`

SamCreasey (approved):

PR #791, opened 2024-03-19
fix(ImageBlock): some svgs dont render

fulopdaniel:
After image processing some SVGs have null for `image.width` and `image.height`.
In these cases render them full width.


ragi96 (approved):
LGTM

PR #804, opened 2024-04-05
feat(ImageBlock): allow attachments when block is empty

fulopdaniel:
https://www.figma.com/file/pYt8ECuB6Sc63cVtpKiDgf/Blocks---Updated?type=design&node-id=485-128059&mode=design&t=Ax3uS4Y06OpRKh6h-0

https://www.loom.com/share/911b7fb9330b4b9a8ac98dd2aae8a03f

CU-86947yy57

https://app.clickup.com/t/86947yy57

## Changes

1. Attachments moved to parent component so overflow:hidden is not applied to that.
2. Added a 44px min height to the image component if there are attachments.
3. Hide captions if there is none in view mode.

### DIFF
--- a/packages/image-block/src/components/ImageCaption.tsx
+++ b/packages/image-block/src/components/ImageCaption.tsx
@@ -9,7 +9,7 @@ import {
 } from '@frontify/guideline-blocks-settings';
 import { getCaptionPlugins, titlePlugins } from './helpers';
 import { AppBridgeBlock } from '@frontify/app-bridge';
-import { CaptionPosition, Ratio, type Settings, textRatioValues } from '../types';
+import { CaptionPosition, type Ratio, type Settings, textRatioValues } from '../types';
 
 type ImageCaptionProps = {
     blockId: string;
@@ -36,7 +36,12 @@ export const ImageCaption = ({
 }: ImageCaptionProps) => {
     const hasTitle = hasRichTextValue(name);
     const hasDescription = hasRichTextValue(description);
-    return isEditing || hasTitle || hasDescription ? (
+
+    if (!isEditing && !hasTitle && !hasDescription) {
+        return null;
+    }
+
+    return (
         <div
             className={
                 positioning === CaptionPosition.Above || positioning === CaptionPosition.Below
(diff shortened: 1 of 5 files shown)


packages/image-block/src/components/ImageCaption.tsx
```diff
@@ -9,7 +9,7 @@ import {
 } from '@frontify/guideline-blocks-settings';
 import { getCaptionPlugins, titlePlugins } from './helpers';
 import { AppBridgeBlock } from '@frontify/app-bridge';
-import { CaptionPosition, type Settings } from '../types';
+import { CaptionPosition, Ratio, type Settings, textRatioValues } from '../types';
```
ragi96, line 12:
```suggestion
import { CaptionPosition, type Ratio, type Settings, textRatioValues } from '../types';
```

ragi96 (commented):
Also add tests at least for these cases:

1. Attachment set - no asset -> Attachment Shown
2. Attachment set - no asset - download active -> Attachment Shown without Download Button

fulopdaniel:
> Screenshot 2024-04-05 at 13 49 01
> 
> Displays the download button without an asset

Nice catch, fixed, thanks! Added a testcase for it. (Merged the two test case suggestions you made since they are essentially the same: hide download, show attachments in view mode.)

packages/image-block/src/components/ImageCaption.tsx
```diff
@@ -30,36 +31,47 @@ export const ImageCaption = ({
     description,
     name,
     positioning,
+    ratio,
     setBlockSettings,
 }: ImageCaptionProps) => {
-    return (
+    const hasTitle = hasRichTextValue(name);
+    const hasDescription = hasRichTextValue(description);
+    return isEditing || hasTitle || hasDescription ? (
```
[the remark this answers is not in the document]
fulopdaniel: Sure, adjusted

packages/image-block/src/components/ImageCaption.tsx
fulopdaniel: fixed

ragi96 (approved):
LGTM

PR #809, opened 2024-04-08
fix: image focus styles

SamCreasey:
Focus styles were hidden by the image wrapper due to an overflow:hidden style being applied. To fix this some refactoring was needed. This also adjusts the image functionality, as now the container of the image will always wrap to the image itself, meaning that if the image is not 100% of the width of the container and you apply a background, it will only be added to the area of padding around the image, instead of the full width, which is also shown in the Figma design to be the correct behaviour

### DIFF
--- a/packages/image-block/src/components/Image.tsx
+++ b/packages/image-block/src/components/Image.tsx
@@ -6,7 +6,7 @@ import { Security, joinClassNames } from '@frontify/guideline-blocks-settings';
 import { ResponsiveImage, useImageContainer } from '@frontify/guideline-blocks-shared';
 import { type CSSProperties, type ReactNode } from 'react';
 
-import { type Link, type Settings, mapAlignmentClasses } from '../types';
+import { Alignment, type Link, type Settings, mapAlignmentClasses } from '../types';
 
 import { getImageWrapperStyle } from './helpers';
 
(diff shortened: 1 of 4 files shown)


fulopdaniel:
Codewise it looks good 👍  However I think it's worth to double check the behavior changes with Martin, as part of the image migration they are super careful everything looks 1 to 1 the same with the old image block. And for the old image block the background color is applied to the whole width. Same for some other styles, e.g. border styles.

packages/image-block/src/components/Image.tsx
```diff
@@ -16,65 +18,87 @@ type ImageProps = {
...
-            <button data-test-id="image-block-asset-viewer-button" {...props} onClick={() => open(image)}>
-                {Image}
+            <button data-test-id="image-block-asset-viewer-wrapper" {...sharedProps} onClick={() => open(image)}>
+                {children}
             </button>
         );
     }
 
-    return Image;
+    return null;
```
fulopdaniel, line 89: nit, to avoid the return null, I would prefer it like this:
```
if (link && !isEditing) return ...
if (isAssetViewerEnabled && !isEditing) return ...
return <div {...sharedProps} {...}>{children}</div>
```

fulopdaniel (approved):

packages/image-block/src/components/Image.tsx
SamCreasey: fixed

PR #816, opened 2024-04-16
fix(ImageBlock): image attachments position is wrong

fulopdaniel:
Relative positioning was missing from the image holder, so attachments were off on some layouts.


fulopdaniel:
CU-86949z1h4

ragi96 (approved):
LGTM

PR #818, opened 2024-04-16
feat(TemplateBlock): template editing setting and CTA action update

ryancarville:
Implements the new `Template Editing` `switch` setting which sets the CTA action to navigate to the new `CreationForm` when `simple` OR use the current template editor when `advanced`.

How to test:
1. `Simple` Editing Mode
  a. navigate to a new `TemplateBlock`
  b. move to `editMode`
  c. open the `settings` sidebar
  d. select `Simple` under the `Template Editing` section (this should be the default)
  e. move back to `viewMode`
  f. `click` the CTA ("Use this Template")
  g. browser should navigate to the `CreationForm` `route` with a `referer` `param`

2. `Advanced` Editing Mode
  a. navigate to a new `TemplateBlock`
  b. move to `editMode`
  c. open the `settings` sidebar
  d. select `Advanced` under the `Template Editing` section
  e. move back to `viewMode`
  f. `click` the CTA ("Use this Template")
  g. should render a `modal` to `CreatePublication`

CU-8694auyrh

### DIFF
--- a/packages/template-block/src/TemplateBlock.tsx
+++ b/packages/template-block/src/TemplateBlock.tsx
@@ -9,7 +9,7 @@ import { AlertError } from './components/AlertError';
 import { CtaButton } from './components/CtaButton';
 import { TemplatePreview } from './components/TemplatePreview';
 import { TemplateText } from './components/TemplateText';
-import { PreviewType } from './types';
+import { PreviewType, TemplateEditing } from './types';
 import { useTemplateBlockData } from './hooks/useTemplateBlockData';
 
 export const TemplateBlock = ({ appBridge }: BlockProps): ReactElement => {
@@ -38,6 +38,15 @@ export const TemplateBlock = ({ appBridge }: BlockProps): ReactElement => {
 
     const handleNewPublication = async () => {
         if (selectedTemplate !== null) {
+            const { templateEditing } = blockSettings;
+            if (templateEditing === TemplateEditing.Simple) {
+                const uri = `/publishing/template/${selectedTemplate.id}?referer=${window.location.href.replaceAll(
+                    '/',
+                    '%2F'
+                )}`;
+                open(uri, '_self');
+                return;
+            }
             const previewUrl =
                 preview === PreviewType.Custom && previewCustom
                     ? previewCustom.previewUrl
--- a/packages/template-block/src/settings.ts
+++ b/packages/template-block/src/settings.ts
@@ -15,12 +15,14 @@ import {
     numericalOrPixelRule,
     presetCustomValue,
     radiusStyleMap,
+    NotificationStyleType,
 } from '@frontify/guideline-blocks-settings';
 import {
     AnchoringType,
     PreviewDisplayType,
     PreviewHeightType,
     PreviewType,
+    TemplateEditing,
     TextPositioningType,
     TextRatioType,
     paddingStyleMap,
@@ -73,6 +75,44 @@ export const settings = defineSettings({
             label: 'Custom Preview',
             show: (bundle: Bundle) => bundle.getBlock('preview')?.value === PreviewType.Custom,
             showForTranslations: true,
+        },
+        {
+            id: 'templateEditingHeading',
+            type: 'sectionHeading',
+            blocks: [
+                {
+                    id: 'templateEditing',
+                    label: 'Template Editing',
+                    info: 'Choose how users can make changes to the template.',
+                    type: 'slider',
+                    choices: [
+                        { label: 'Simple', value: TemplateEditing.Simple },
+                        { label: 'Advanced', value: TemplateEditing.Advanced },
+                    ],
+                    defaultValue: TemplateEditing.Simple,
+                    showForTranslations: true,
+                },
+                {
+                    id: 'templateEditingNotificationSimple',
+                    type: 'notification',
+                    text: 'Users can only make changes by selecting and editing fields in the settings panel without approvals. Read <article name> to learn more.',
+                    show: (bundle: Bundle) => bundle.getBlock('templateEditing')?.value === TemplateEditing.Simple,
+                    styles: {
+                        type: NotificationStyleType.Info,
+                    },
+                    showForTranslations: true,
+                },
+                {
+                    id: 'templateEditingNotificationAdvanced',
+                    type: 'notification',
+                    text: 'Users can select layers and edit text directly. Read <article name> to learn more.',
+                    show: (bundle: Bundle) => bundle.getBlock('templateEditing')?.value === TemplateEditing.Advanced,
+                    styles: {
+                        type: NotificationStyleType.Info,
+                    },
+                    showForTranslations: true,
+                },
+            ],
         },
     ],
     layout: [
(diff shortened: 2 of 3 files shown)


packages/template-block/src/TemplateBlock.tsx
```diff
@@ -38,6 +38,15 @@ export const TemplateBlock = ({ appBridge }: BlockProps): ReactElement => {
 
     const handleNewPublication = async () => {
         if (selectedTemplate !== null) {
+            const { templateEditing } = blockSettings;
+            if (templateEditing === TemplateEditing.Simple) {
+                const uri = `/publishing/template/${selectedTemplate.id}?referer=${window.location.href.replaceAll(
```
mikeyrayvon, line 43: I would prefer `encodeURI()` here

```suggestion
                const uri = `/publishing/template/${selectedTemplate.id}?referer=${encodeURI(window.location.href)}`
```

packages/template-block/src/settings.ts
```diff
@@ -74,6 +76,44 @@ export const settings = defineSettings({
...
+                        { label: 'Simple', value: TemplateEditing.Simple },
+                        { label: 'Advanced', value: TemplateEditing.Advanced },
+                    ],
+                    defaultValue: TemplateEditing.Simple,
+                    showForTranslations: true,
+                },
+                {
+                    id: 'templateEditingNotificationSimple',
+                    type: 'notification',
+                    text: 'Users can only make changes by selecting and editing fields in the settings panel without approvals. Read <article name> to learn more.',
```
mikeyrayvon, line 98: Is it even possible to have a link here in place of `<articleName>`?

mikeyrayvon: But generally, I would think better to leave out the `Read <article name> to learn more.` entirely until we have an actual article to put there

mikeyrayvon (changes requested):

packages/template-block/src/settings.ts
```diff
@@ -74,6 +76,44 @@ export const settings = defineSettings({
...
+                    show: (bundle: Bundle) => bundle.getBlock('templateEditing')?.value === TemplateEditing.Simple,
+                    styles: {
+                        type: NotificationStyleType.Info,
+                    },
+                    showForTranslations: true,
+                },
+                {
+                    id: 'templateEditingNotificationAdvanced',
+                    type: 'notification',
+                    text: 'Users can select layers and edit text directly. Read <article name> to learn more.',
```
mikeyrayvon, line 108: Same comments here

packages/template-block/src/settings.ts
ryancarville: Would be possible to use a link by moving this to a `footer` section of the notifications element I believe.  But we can revisit this once we have copy

mikeyrayvon (approved):
Works great
Nice code
💯

packages/template-block/src/TemplateBlock.tsx
```diff
@@ -38,6 +38,12 @@ export const TemplateBlock = ({ appBridge }: BlockProps): ReactElement => {
 
     const handleNewPublication = async () => {
         if (selectedTemplate !== null) {
+            const { templateEditing } = blockSettings;
+            if (templateEditing === TemplateEditing.Simple) {
+                const uri = `/publishing/template/${selectedTemplate.id}?referer=${encodeURI(window.location.href)}`;
+                open(uri, '_self');
+                return;
+            }
```
mikeyrayvon, line 46:
```suggestion
            }

```

packages/template-block/src/TemplateBlock.tsx
```diff
@@ -38,6 +38,12 @@ export const TemplateBlock = ({ appBridge }: BlockProps): ReactElement => {
 
     const handleNewPublication = async () => {
         if (selectedTemplate !== null) {
+            const { templateEditing } = blockSettings;
```
mikeyrayvon, line 41:
```suggestion
            const { templateEditing } = blockSettings;

```

packages/template-block/src/TemplateBlock.tsx
```diff
@@ -38,6 +38,12 @@ export const TemplateBlock = ({ appBridge }: BlockProps): ReactElement => {
 
     const handleNewPublication = async () => {
         if (selectedTemplate !== null) {
```
mikeyrayvon, line 40: Ah also, lets invert this for an early return @ryancarville 

```suggestion
        if (!selectedTemplate) {
        	return;
        }
```

mikeyrayvon (changes requested):

packages/template-block/src/TemplateBlock.tsx
mikeyrayvon: 🌬️

packages/template-block/src/TemplateBlock.tsx
mikeyrayvon: 🌬️

mikeyrayvon (approved):

packages/template-block/src/TemplateBlock.tsx
```diff
@@ -37,21 +37,26 @@ export const TemplateBlock = ({ appBridge }: BlockProps): ReactElement => {
...
-                },
-            };
+        if (!selectedTemplate) {
+            return;
+        }
+        const { templateEditing } = blockSettings;
 
-            await appBridge.dispatch(openNewPublication(options));
+        if (templateEditing === TemplateEditing.Simple) {
+            const uri = `/publishing/template/${selectedTemplate.id}?referer=${encodeURI(window.location.href)}`;
```
ragi96, line 46: imo this `uri` should go into the `Template` type within the `app-bridge`.
So you have full control over it via the web-app, if  something changes you can change it without touching this block (or even blocks of partner if they build there own)

ragi96: As discussed this uri will move into the template type

packages/template-block/src/TemplateBlock.tsx
```diff
@@ -9,7 +9,7 @@ import { AlertError } from './components/AlertError';
 import { CtaButton } from './components/CtaButton';
```
ragi96, line 9: Would love to have a test for the adjustments, otherwise its fine for me

packages/template-block/src/TemplateBlock.spec.ct.tsx
```diff
@@ -638,4 +638,19 @@ describe('Template Block', () => {
...
+    it('should set CTA path to CreationFormUri', () => {
+        const [TemplateBlockWithStubs] = withAppBridgeBlockStubs(TemplateBlock, {
+            editorState: false,
+            blockTemplates: {
+                template: [getTemplateDummyWithPages()],
+            },
+        });
+
+        mount(<TemplateBlockWithStubs />);
+        cy.get(CTA_BUTTON_SELECTOR).click({ force: true });
```
ragi96, line 651:
```suggestion
        cy.get(CTA_BUTTON_SELECTOR).realClick();
```

ragi96 (approved):

PR #848, opened 2024-05-14
feat(ImageBlock): set background color to background color of the asset

ragi96:
Last part of [Task](https://app.clickup.com/t/2523021/TASK-11147)

Needs first released: https://github.com/Frontify/web-app/pull/4132


packages/image-block/src/ImageBlock.tsx
```diff
@@ -74,11 +75,19 @@ export const ImageBlock = withAttachmentsProvider(({ appBridge }: BlockProps) =>
...
+                settings.name = convertToRteValue(TextStyles.imageTitle, defaultImageName, 'center');
                 setTitleKey(generateRandomId());
             }
         }
 
+        if (newImage.backgroundColor) {
+            settings.backgroundColor = newImage.backgroundColor;
+            settings.hasBackground = true;
+        } else {
+            settings.hasBackground = false;
```
fulopdaniel, line 87: In this scenario should we just keep the existing setting, instead of setting it to false?

ragi96: Its cloning what the old image block did, if you added a first image with background color and then you add one without it has remove the background color

ragi96: Quickly discussed it with martin

and we will remove the removage of the background color.

fulopdaniel (approved):

PR #849, opened 2024-05-15
feat(ImageBlock): respect downloadProtection for the downloadButton

ragi96:
Needs https://github.com/Frontify/web-app/pull/4158


packages/image-block/src/components/DownloadAndAttachments.tsx
```diff
@@ -35,9 +35,12 @@ export const DownloadAndAttachments = ({
                 data-test-id="buttons-wrapper"
                 style={getTotalImagePadding(blockSettings)}
             >
-                {image && isDownloadable(blockSettings.security, blockSettings.downloadable, assetDownloadEnabled) && (
-                    <DownloadButton onDownload={() => appBridge.dispatch({ name: 'downloadAsset', payload: image })} />
-                )}
+                {image?.isDownloadProtected === false &&
```
fulopdaniel, line 38: Is there a scenario where `image?.isDownloadProtected` is undefined?

ragi96: If I adjust it, we would get a typecheck error on the `dispatch` of the `download`

fulopdaniel (approved):

PR #850, opened 2024-05-15
feat(ImageBlock): respect downloadability in asset viewer

fulopdaniel:


packages/image-block/src/ImageBlock.tsx
```diff
@@ -56,6 +57,12 @@ export const ImageBlock = withAttachmentsProvider(({ appBridge }: BlockProps) =>
     const { blockAssets, deleteAssetIdsFromKey, updateAssetIdsFromKey } = useBlockAssets(appBridge);
     const image = blockAssets?.[IMAGE_ID]?.[0];
     const attachmentCount = blockAssets[ATTACHMENTS_ASSET_ID]?.length || 0;
+    const isImageDownloadable = isDownloadable(
```
ragi96, line 60: add here also the check if its a protected image, 

If its protected always false, otherewise determine based on the `isDownloadable`

We will move that logic in the long run into the `guideline-blocks-settings` but lets wait, so its not becoming blocking

fulopdaniel: Yep, updated!

packages/image-block/src/ImageBlock.tsx
```diff
@@ -56,6 +57,10 @@ export const ImageBlock = withAttachmentsProvider(({ appBridge }: BlockProps) =>
     const { blockAssets, deleteAssetIdsFromKey, updateAssetIdsFromKey } = useBlockAssets(appBridge);
     const image = blockAssets?.[IMAGE_ID]?.[0];
     const attachmentCount = blockAssets[ATTACHMENTS_ASSET_ID]?.length || 0;
+    const isImageDownloadable =
+        image?.isDownloadProtected === false &&
+        isDownloadable(blockSettings.security, blockSettings.downloadable, assetDownloadEnabled);
```
ragi96, line 62: I think it would make sense to move that into the helpers and add some unit tests for it

ragi96: And as i just saw it the helper that is already there, should move into the `shared`

fulopdaniel: Updated, and added some tests to test the new logic, as the isDownloadable is already tested in `guideline-blocks-settings`

ragi96 (approved):
LGTM

PR #854, opened 2024-05-22
fix(ThumbnailGridBlock): only store items if isEditing

ragi96:
Just saw that error: https://sentry.appsupport.frontify.dev/organizations/frontify/issues/131754/?project=3&referrer=slack
This should be prevented if we only store if the user is edit mode, wdyt?

### DIFF
--- a/packages/thumbnail-grid-block/src/ThumbnailGridBlock.tsx
+++ b/packages/thumbnail-grid-block/src/ThumbnailGridBlock.tsx
@@ -80,9 +80,11 @@ export const ThumbnailGridBlock = ({ appBridge }: BlockProps) => {
     };
 
     useEffect(() => {
-        setBlockSettings({ items: itemsState });
+        if (isEditing) {
+            setBlockSettings({ items: itemsState });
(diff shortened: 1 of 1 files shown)


packages/thumbnail-grid-block/src/ThumbnailGridBlock.tsx
```diff
@@ -80,9 +80,11 @@ export const ThumbnailGridBlock = ({ appBridge }: BlockProps) => {
     };
 
     useEffect(() => {
-        setBlockSettings({ items: itemsState });
+        if (isEditing) {
+            setBlockSettings({ items: itemsState });
+        }
         // eslint-disable-next-line react-hooks/exhaustive-deps
-    }, [itemsState]);
+    }, [itemsState, isEditing]);
```
fulopdaniel, line 87: This will fire an api request every time the user switches to edit mode.

What we could do is introduce a variable like isMounted, and prevent this useeffect on its first run, as I suspect that is where the error occurs in view mode.

PR #862, opened 2024-06-07
feat(ImageBlock): update breakpoints to container breakpoints

fulopdaniel:
Discussed with design that the correct new breakpoint is `sm` instead of `md`.

Most of the changes in `ImageBlock.tsx` are just whitespace changes. I added a wrapper div around.

### DIFF
--- a/packages/image-block/src/ImageBlock.tsx
+++ b/packages/image-block/src/ImageBlock.tsx
@@ -142,113 +142,118 @@ export const ImageBlock = withAttachmentsProvider(({ appBridge }: BlockProps) =>
 
     return (
         <div className="image-block">
-            <div data-test-id="image-block" className={`tw-flex tw-h-auto ${mapCaptionPositionClasses[positioning]}`}>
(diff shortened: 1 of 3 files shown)


packages/image-block/src/ImageBlock.tsx
```diff
@@ -142,113 +142,118 @@ export const ImageBlock = withAttachmentsProvider(({ appBridge }: BlockProps) =>
 
     return (
         <div className="image-block">
-            <div data-test-id="image-block" className={`tw-flex tw-h-auto ${mapCaptionPositionClasses[positioning]}`}>
+            <div className="tw-@container">
```
ragi96, line 145: Why an extra div here? can't we move the `tw-@container` to the scoping div which is around?

fulopdaniel: because of how our postcss plugin works, that scopes every css generated to be inside `.image-block`

ragi96 (approved):

PR #880, opened 2024-06-27
feat(CodeSandboxBlock): add static support

fulopdaniel:
https://app.clickup.com/t/8694mhxzg

### DIFF
--- a/packages/ui-pattern-block/src/components/Captions.tsx
+++ b/packages/ui-pattern-block/src/components/Captions.tsx
@@ -34,7 +34,7 @@ export const Captions = ({
     };
 
     return (
-        <div className={joinClassNames(['tw-flex tw-flex-col tw-gap-1'])}>
+        <div className={joinClassNames(['tw-flex tw-flex-col tw-gap-1 tw-w-full'])}>
             <RichTextEditor
                 id={`${appBridge.getBlockId().toString()}_title`}
                 value={title ?? convertToRteValue(TextStyles.imageTitle, '')}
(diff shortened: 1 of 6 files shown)


packages/ui-pattern-block/src/components/Captions.tsx
```diff
@@ -34,7 +34,7 @@ export const Captions = ({
     };
 
     return (
-        <div className={joinClassNames(['tw-flex tw-flex-col tw-gap-1'])}>
+        <div className={joinClassNames(['tw-flex tw-flex-col tw-gap-1 tw-w-full'])}>
```
ragi96, line 37: Why is this `tw-w-full` needed

ragi96 (changes requested):
External Deps and NPM deps not working, so we should disable that field then

packages/ui-pattern-block/src/components/Captions.tsx
fulopdaniel: Otherwise this happens now. Probably some change around the scoping/fondue changes caused this.

image

fulopdaniel:
> Member

good call, fixed

packages/ui-pattern-block/src/components/Captions.tsx
ragi96: Ok fine

packages/ui-pattern-block/src/helpers/files.ts
```diff
@@ -45,6 +45,22 @@ export const getDefaultFilesOfTemplate = (template: SandpackTemplate, preprocess
...
+<html>
+<head>
+  <script src='./index.js'></script>
+  <link href="style.css" rel="stylesheet" />
+</head>
+<body>
+  <h1>Hello world</h1>
+</body>
+</html>`,
+                '/index.js': 'console.log("Mounted")',
```
ragi96, line 60:
```suggestion
                '/index.js': 'console.log("Sandbox Ready")',
```

ragi96 (commented):
Much better now, thanks :)

ragi96 (approved):

PR #893, opened 2024-07-16
fix(ThumbnailGridBlock): drag and drop is not working and laggy

fulopdaniel:
https://app.clickup.com/t/86952t6yy

Had to move SortableContext outside the `tw-@container` definition.

I also improved performance greatly by stabilising the updateItem prop, and memoising the RTE component.


ragi96 (approved):
Feels great now :)

PR #895, opened 2024-07-19
fix(CalloutBlock): hug contents is collapsing

fulopdaniel:
[From MDN docs:](https://developer.mozilla.org/en-US/docs/Web/CSS/container-type)

`Size containment turns off the ability of an element to get size information from its contents, which is important for container queries to avoid infinite loops.`

Meaning with container queries it is not possible to get the width of the content inside, so hugging the content width is not possible.

### DIFF
--- a/packages/callout-block/src/CalloutBlock.spec.ct.tsx
+++ b/packages/callout-block/src/CalloutBlock.spec.ct.tsx
@@ -281,7 +281,8 @@ describe('Callout Block', () => {
 
         cy.get(CalloutBlockSelector).should(($el) => {
             const width = parseFloat($el.css('width'));
-            expect(width).to.be.greaterThan(100);
+            expect(width).to.be.greaterThan(130);
+            expect(width).to.be.lessThan(140);
         });
     });
 });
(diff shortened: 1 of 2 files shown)


ragi96:
Can we add a test to avoid regression? Something like adding content to a callout with hugWidth and then the width of the block should be XXpx?

ragi96 (commented):
Works but as already stated a test to avoid regression would be good

fulopdaniel:
> Works but as already stated a test to avoid regression would be good

Yep, added one

packages/callout-block/src/CalloutBlock.spec.ct.tsx
```diff
@@ -266,4 +266,22 @@ describe('Callout Block', () => {
...
+                appearance: Appearance.Strong,
+                width: Width.HugContents,
+            },
+        });
+
+        mount(<CalloutBlockWithStubs />);
+
+        cy.get(CalloutBlockSelector).should(($el) => {
+            const width = parseFloat($el.css('width'));
+            expect(width).to.be.greaterThan(100);
```
ragi96, line 284: can we do this more exactly as its also true if its full width what also would be wrong

ragi96 (approved):

ragi96:
Will merge and release it :)

PR #904, opened 2024-07-31
fix(PostCSS): only scope classes

fulopdaniel:
Only extend the scoping to modals if the selector contains a class.

This way selectors like `*` `:before` `h1` will not be added to modals and overlays which previously overrode things.


anxobotana (approved):
LGTM

PR #911, opened 2024-08-06
fix(DoDontBlock): initial color saving runs into race condition

fulopdaniel:
Previously 2 requests were fired at the same time => race condition.

Merged them now into 1 request.


Kenny806 (approved):
LGTM

PR #929, opened 2024-09-10
fix(FigmaBlock): asset chooser does not open

fulopdaniel:
Move `<FigmaEmptyBlock />` to a new component. Otherwise the click handler was never called on the button.

I suspect due to it being recreated on every rerender and the mousedown => mouseup was happening on two different components.


ragi96 (approved):
I love it (there is so much more to clean this block up 😢 )

Deployed v21 and did a private release for testing on product, seems working again. Thanks 👍 

Lets ship it

PR #975, opened 2024-12-05
feat(QuoteBlock): assets do not unregister on switch off

fulopdaniel:


packages/quote-block/src/QuoteBlock.tsx
```diff
@@ -45,6 +45,25 @@ export const QuoteBlock: FC<BlockProps> = ({ appBridge }) => {
     const [blockSettings, setBlockSettings] = useBlockSettings<Settings>(appBridge);
     const isEditing = useEditorState(appBridge);
 
+    const { blockAssets, deleteAssetIdsFromKey } = useBlockAssets(appBridge);
+    const customLeftIcon = blockAssets?.[CUSTOM_QUOTE_STYLE_LEFT_ID]?.[0];
+    const customRightIcon = blockAssets?.[CUSTOM_QUOTE_STYLE_RIGHT_ID]?.[0];
+
+    useEffect(() => {
```
ragi96, line 52: We should check if `isEditing` in here, to not call the api with calls that will be aborted...

PR #986, opened 2025-01-08
fix(DoDont): overlapping and background color

ragi96:
Issue found during testing:
- background color of block (removed)
- overlapping outline of the blockItemWrapper (offset to 0)

Also replaced deprecated `getBlockId()`

Also found an issue with the scoping around tooltips, as the class was prefixed but lived on the root. 
So now we only prefix tailwind classes and tags and no other classes this fixes the issue bolding without introducing new issues. 

https://app.clickup.com/t/2523021/TASK-13279


silviojaeger (approved):
LGTM👍

PR #990, opened 2025-01-09
feat(DnDBlock): implement column reflow based on container size

peter-tudosa:
Ticket: CU-86950mpmn

### DIFF
--- a/cypress/support/component-index.html
+++ b/cypress/support/component-index.html
@@ -1,15 +1,12 @@
 <!DOCTYPE html>
 <html xml:lang="en">
-
-<head>
-    <meta charset="utf-8" />
-    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
-    <meta name="viewport" content="width=device-width,initial-scale=1.0" />
-    <title>Components App</title>
-</head>
-
-<body>
-    <div data-cy-root></div>
-</body>
-
+    <head>
+        <meta charset="utf-8" />
+        <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+        <meta name="viewport" content="width=device-width,initial-scale=1.0" />
+        <title>Components App</title>
+    </head>
+    <body>
+        <div data-cy-root></div>
+    </body>
 </html>
--- a/cypress/support/component-index.html
+++ b/cypress/support/component-index.html
@@ -6,11 +6,6 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
     <meta name="viewport" content="width=device-width,initial-scale=1.0" />
     <title>Components App</title>
-    <style>
-        @tailwind base;
-        @tailwind components;
-        @tailwind utilities;
-    </style>
 </head>
 
 <body>
--- a/cypress/support/component-index.html
+++ b/cypress/support/component-index.html
@@ -1,12 +1,20 @@
 <!DOCTYPE html>
 <html xml:lang="en">
-    <head>
-        <meta charset="utf-8" />
-        <meta http-equiv="X-UA-Compatible" content="IE=edge" />
-        <meta name="viewport" content="width=device-width,initial-scale=1.0" />
-        <title>Components App</title>
-    </head>
-    <body>
-        <div data-cy-root></div>
-    </body>
+
+<head>
+    <meta charset="utf-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width,initial-scale=1.0" />
+    <title>Components App</title>
+    <style>
+        @tailwind base;
+        @tailwind components;
+        @tailwind utilities;
+    </style>
+</head>
+
+<body>
+    <div data-cy-root></div>
+</body>
+
 </html>
--- a/packages/dos-donts-block/src/DosDontsBlock.spec.ct.tsx
+++ b/packages/dos-donts-block/src/DosDontsBlock.spec.ct.tsx
@@ -199,11 +199,9 @@ describe("Dos & Don'ts Block", () => {
                 },
             });
 
-            mount(
-                <div className="tw-w-[150px]">
-                    <DosDontsBlockWithStubs />
-                </div>
-            );
+            cy.viewport(150, 500);
+
+            mount(<DosDontsBlockWithStubs />);
 
             cy.get(DND_ADD_BUTTONS).should('have.css', 'flex-direction', 'column');
         });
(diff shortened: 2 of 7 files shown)


ragi96 (commented):
Reflow is fine but imo we also need to care about the blank state:

Screenshot 2025-01-09 at 17 31 06

Also breaking to 1 col should be considered (atm not there), needs quick check with lucija
Screenshot 2025-01-09 at 17 32 32

cypress/support/component-index.html
```diff
@@ -1,12 +1,15 @@
 <!DOCTYPE html>
```
ragi96, line 1: No change in here, besides some spaces?

packages/dos-donts-block/src/DosDontsBlock.spec.ct.tsx
```diff
@@ -186,4 +187,44 @@ describe("Dos & Don'ts Block", () => {
...
+                editorState: true,
+                blockSettings: {
+                    style: DoDontStyle.Icons,
+                    dontIconChoice: ItemIconChoice.CHECKMARK,
+                    doIconChoice: ItemIconChoice.CHECKMARK,
+                },
+            });
+
+            mount(
+                <div className="tw-w-[500px]">
```
ragi96, line 222: Wouldn't wrap here but just use:

```typescript
cy.viewport()
``` 

Same for the other

packages/dos-donts-block/src/DosDontsBlock.spec.ct.tsx
```diff
@@ -186,4 +187,44 @@ describe("Dos & Don'ts Block", () => {
         cy.get(FLOATING_BUTTON_BUTTON).should('exist').click();
         cy.get(INTERNAL_LINK_SELECTOR).should('exist');
     });
+
+    describe('Add buttons', () => {
```
ragi96, line 191: no need for this subscribe

packages/dos-donts-block/src/DosDontsBlock.tsx
```diff
@@ -188,6 +188,34 @@ export const DosDontsBlock: FC<BlockProps> = ({ appBridge }) => {
...
+
+        const observer = new ResizeObserver((entries) => {
+            if (entries[entries.length - 1].contentRect.width < CONTAINER_SMALL_LIMIT) {
+                setisContainerSmall(true);
+            } else {
+                setisContainerSmall(false);
+            }
+        });
+
+        observer.observe(container);
```
ragi96, line 210: throttle here

packages/dos-donts-block/src/DosDontsBlock.tsx
```diff
@@ -188,6 +188,34 @@ export const DosDontsBlock: FC<BlockProps> = ({ appBridge }) => {
         // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [doneAll, uploadResults]);
 
+    const containerRef = useRef<HTMLDivElement>(null);
+    const [isContainerSmall, setisContainerSmall] = useState<boolean>(false);
```
ragi96, line 192: move them to the top

packages/dos-donts-block/src/DosDontsBlock.tsx
```diff
@@ -188,6 +188,34 @@ export const DosDontsBlock: FC<BlockProps> = ({ appBridge }) => {
...
+    const CONTAINER_SMALL_LIMIT = 440; // this is the value defined in tailwind for @sm
```
ragi96, line 193: move this one to the `const.ts`

packages/dos-donts-block/src/DosDontsBlock.tsx
```diff
@@ -188,6 +188,34 @@ export const DosDontsBlock: FC<BlockProps> = ({ appBridge }) => {
...
+        if (!container) {
+            return;
+        }
+
+        const observer = new ResizeObserver((entries) => {
+            if (entries[entries.length - 1].contentRect.width < CONTAINER_SMALL_LIMIT) {
+                setisContainerSmall(true);
+            } else {
+                setisContainerSmall(false);
+            }
```
ragi96, line 207:
```suggestion
    const lastEntry = entries[entries.length - 1];
    const isSmall = lastEntry?.contentRect?.width < CONTAINER_SMALL_LIMIT;
    setIsContainerSmall(isSmall);
```

better readability

packages/dos-donts-block/src/DosDontsBlock.tsx
```diff
@@ -420,14 +448,19 @@ export const DosDontsBlock: FC<BlockProps> = ({ appBridge }) => {
                             />
                         </div>
                     )}
-                    <div className="tw-flex tw-w-full">
+                    <div
+                        data-test-id="dos-donts-block-add-buttons"
+                        className="tw-flex tw-w-full tw-flex-col @sm:tw-flex-row"
+                    >
                         <BlockInjectButton
+                            verticalLayout={isContainerSmall}
```
ragi96, line 456: In the long run we need to adjust that component so it cares by itself about the vertical layout, based on css? wdyt?

peter-tudosa: Would be sweet but I'm not sure about what implications does this have. This would basically require some modifications in the source code of the components, where are these buttons used elsewhere, do we want the same behavior for all the other usages?

cypress/support/component-index.html
peter-tudosa: I fiddled with it at one point and then reverted, I must've left something in there.

packages/dos-donts-block/src/DosDontsBlock.tsx
ragi96: They are here https://github.com/Frontify/brand-sdk/blob/main/packages/guideline-blocks-settings/src/components/BlockInjectButton/BlockInjectButton.tsx

Our usage is within this guideline-blocks repo, outside usage i am not sure, but we shouldn't fear to change it. We can do breaking changes (could also be a non breaking)

packages/dos-donts-block/src/DosDontsBlock.spec.ct.tsx
peter-tudosa: Hm... I kinda like how the tests read with the label.

image

I feel like without it doesn't make that much sense.

cypress/support/component-index.html
peter-tudosa: reverted it!

ragi96 (approved):
LGTM

PR #996, opened 2025-01-20
fix(QuoteBlock): prevent text from overflowing outside container

peter-tudosa:
Ticket: CU-8697g24cc

Before:
image
image

After:
image
image


fulopdaniel:
Awesome fix. Two issue I found is the author is sometimes broken into two lines, and sometimes it is cut off.

image
image

peter-tudosa:
> Awesome fix. Two issue I found is the author is sometimes broken into two lines, and sometimes it is cut off.
> 
> image image

@fulopdaniel  I added a `break-all` on the author text to prevent that weird cutout of text.

fulopdaniel (approved):
nice

PR #999, opened 2025-01-23
fix(PostCSS): scope root vars to block

fulopdaniel:
This way root css vars bundled within the block do not interfere with web app vars.


ragi96 (approved):
LGTM

PR #1000, opened 2025-01-23
chore(*): remove unused blocks

fulopdaniel:
Removed:

```
brand-positioning-block
checklist-block
color-block
color-kit-block
color-scale-block
personal-note-block
```


ragi96 (approved):
Great that its number 1000, thanks for taking care 🚀

PR #1006, opened 2025-01-27
fix(DnDBlock): button stacking on small sized containers

peter-tudosa:
Ticket: CU-8697n0fc9

# Before:
image

# After: 
image

### DIFF
--- a/packages/dos-donts-block/src/DosDontsBlock.tsx
+++ b/packages/dos-donts-block/src/DosDontsBlock.tsx
@@ -434,7 +434,7 @@ export const DosDontsBlock: FC<BlockProps> = ({ appBridge }) => {
                 ) : null}
             </DndContext>
             {isEditing && (
-                <div className="tw-w-full tw-flex tw-gap-3 tw-mt-9">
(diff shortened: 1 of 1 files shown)


packages/dos-donts-block/src/DosDontsBlock.tsx
```diff
@@ -434,7 +434,7 @@ export const DosDontsBlock: FC<BlockProps> = ({ appBridge }) => {
                 ) : null}
             </DndContext>
             {isEditing && (
-                <div className="tw-w-full tw-flex tw-gap-3 tw-mt-9">
+                <div className="tw-w-full tw-flex tw-gap-3 tw-mt-9 tw-flex-col @md:tw-flex-row">
```
fulopdaniel, line 437: question: does `flex-wrap` not work in this scenario, maybe in pair with `flex-shrink-0` on children?

peter-tudosa: It seems like it's working with `flex-wrap`, thanks for the suggestion! 🤘

PR #1008, opened 2025-01-27
fix(ThumbnailGridBlock):  remove unnecessary css property

peter-tudosa:
Remove word-break style that was preventing the buttons to work properly


ragi96 (approved):
LGTM

PR #1011, opened 2025-02-04
chore: bump fondue to 12.5.1

fulopdaniel:
Created a [StyleProvider](https://github.com/Frontify/guideline-blocks/pull/1011/files#diff-3dcf25ca7887e80de0b985f4ab63c4788c0beae8cf74aa8a4be2914cffde8e22) component in the `shared` folder that provides the theme and other necessary CSS imports.


ragi96:
Imo you need to do it first within the guideline-blocks-settings, not?

fulopdaniel:
> Imo you need to do it first within the guideline-blocks-settings, not?

Nope, this picks up the correct version. Tried with a private deployment and it works.

examples/asset-upload/package.json
SamuelAlev, line 1: As part of this PR we added a Theme Provider which scope the CSS variables: https://fondue-components.frontify.com/?path=/docs/1-getting-started--documentation#tokens--themes-new-in-v124

Please wrap all the block with `<ThemeProvider theme="light">` (from `@frontify/fondue/tokens`) and add the scoped styles (`import "@frontify/fondue/tokens/themes";`) and then you can remove every other styles import

SamuelAlev: It is not a breaking change, but better do it ASAP so you have less side-effect with the global CSS on `:root`

ragi96: Great, thanks 🙌

SamuelAlev: CC'ing @noahwaldner as he is playing around it currently.

noahwaldner: I willmping you when i finish moving the provider

noahwaldner: @fulopdaniel Version 12.5.1 is now available and you should be able to import ThemeProvider from @frontif/fondue/components 

Example usages cam be found in web-app

let me know if you need help adapting it

packages/shared/src/components/StyleProvider/index.tsx
```diff
@@ -0,0 +1,3 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+export * from './StyleProvider';
```
noahwaldner, line 3: Please do not use Barrel exports. 
Using Barrel exports make bundling way harder and are known to cause performance issues and circular dependencies.

Some resources on the topic:
https://tkdodo.eu/blog/please-stop-using-barrel-files
https://dev.to/tassiofront/barrel-files-and-why-you-should-stop-using-them-now-bc4

packages/shared/src/components/index.ts
```diff
@@ -3,3 +3,4 @@
 export * from './CssValueDisplay';
 export * from './EditAltTextFlyout';
 export * from './ResponsiveImage';
+export * from './StyleProvider';
```
noahwaldner, line 6: Avoid barrel exports. See comment above

packages/animation-curve-block/src/AnimationCurveBlock.tsx
```diff
@@ -4,16 +4,14 @@ import { useState } from 'react';
 import { DndContext, DragEndEvent, DragOverlay, DragStartEvent, closestCenter } from '@dnd-kit/core';
 import { restrictToParentElement } from '@dnd-kit/modifiers';
 import { SortableContext, arrayMove, rectSortingStrategy } from '@dnd-kit/sortable';
-import '@frontify/guideline-blocks-settings/styles';
```
ragi96, line 7: Sure we can also remove this one? Because of CQ etc?

ragi96 (approved):
LGTM

PR #1025, opened 2025-03-03
refactor(Audio): remove CC

ragi96:
After discussion of it, this is useless on Audio Block and only needed for video. 

For Audio we need transscript which is a different concept, and needs to be checked on how we gonna implement it.


fulopdaniel (approved):
makes sense

PR #1026, opened 2025-03-03
feat(TextBlock): add a min height of 36px

peter-tudosa:
Design requires a min height of 36px for the Text component.

### DIFF
--- a/packages/text-block/src/TextBlock.spec.ct.tsx
+++ b/packages/text-block/src/TextBlock.spec.ct.tsx
@@ -71,4 +71,14 @@ describe('Text Block', () => {
         mount(<TextBlockWithStubs />);
         cy.get(TextBlockSelectorHtml).should('have.text', 'string');
     });
+
+    it('should have a minimum height of 36px', () => {
+        const [TextBlockWithStubs] = withAppBridgeBlockStubs(TextBlock, {
+            editorState: true,
+        });
+
+        mount(<TextBlockWithStubs />);
+
+        cy.get('.text-block').should('have.css', 'min-height', '36px');
+    });
 });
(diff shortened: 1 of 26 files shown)


fulopdaniel:
Task linked: [CU-86984ne9z Add min. height to Text block in Edit mode](https://app.clickup.com/t/86984ne9z)

packages/text-block/src/TextBlock.tsx
```diff
@@ -21,7 +21,7 @@ export const TextBlock = ({ appBridge }: BlockProps): ReactElement => {
     const handleTextChange = useCallback((content: string) => setBlockSettings({ content }), [setBlockSettings]);
 
     return (
-        <div className="text-block">
+        <div className="text-block tw-min-h-9">
```
ragi96, line 24: Imo only in edit mode

ragi96 (changes requested):

packages/text-block/src/TextBlock.tsx
peter-tudosa: Totally right! Srry I missed that.

packages/text-block/src/TextBlock.spec.ct.tsx
```diff
@@ -71,4 +71,24 @@ describe('Text Block', () => {
...
+    });
+
+    it('should not have a minimum height of 36px when editing mode is of', () => {
+        const [TextBlockWithStubs] = withAppBridgeBlockStubs(TextBlock, {
+            editorState: false,
+        });
+
+        mount(<TextBlockWithStubs />);
+
+        cy.get('.text-block').should('not.have.css', 'min-height', '36px');
```
ragi96, line 92: use `data-test-id` here, add one if we don't have one ;)

peter-tudosa: sure

ragi96 (approved):
LGTM

PR #1028, opened 2025-03-07
refactor(AssetKit): remove save download url

ragi96:
As its cached anyways in the backend, I remove this cache. 

Also it tried to store it even it didn't right to do it (Viewer)


fulopdaniel (approved):
lgtm

PR #1032, opened 2025-03-11
fix(Quote): author color should be body

ragi96:
LBD: https://frontify.slack.com/archives/CCXT7AFD0/p1741703226405329

### DIFF
--- a/packages/quote-block/src/QuoteBlock.tsx
+++ b/packages/quote-block/src/QuoteBlock.tsx
@@ -151,7 +151,12 @@ export const QuoteBlock: FC<BlockProps> = ({ appBridge }) => {
                                 />
                             </div>
                             {showAuthor && (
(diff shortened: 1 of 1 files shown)


packages/quote-block/src/QuoteBlock.tsx
```diff
@@ -151,7 +151,12 @@ export const QuoteBlock: FC<BlockProps> = ({ appBridge }) => {
                                 />
                             </div>
                             {showAuthor && (
-                                <p className="tw-text-right tw-break-all">{`- ${blockSettings.authorName}`}</p>
+                                <p
+                                    className="tw-text-right tw-break-all"
+                                    style={{
+                                        color: `var(${THEME_PREFIX}body-color)`,
```
hochreutenerl, line 157: Is this really the way to go? Looks ugly.

fulopdaniel (approved):

packages/quote-block/src/QuoteBlock.tsx
ragi96: indeed, forgot about that one :D

PR #1033, opened 2025-03-13
feat(Callout): use `MutationObserver` instead of `HubAppearanceUpdated` event listener

oliverschwendener:
Instead of listening for the `HubAppearanceUpdated` event we can just observe the style tag that contains all CSS variables, to update the styles of the callout block. This makes sure that the callout block is compatible with the new design settings, which don't emit the `HubAppearanceUpdated` event anymore when changed.

CU-86989mrb0

### DIFF
--- a/packages/callout-block/src/CalloutBlock.tsx
+++ b/packages/callout-block/src/CalloutBlock.tsx
@@ -30,6 +30,9 @@ export const CalloutBlock = ({ appBridge }: BlockProps): ReactElement => {
     const { blockAssets, deleteAssetIdsFromKey } = useBlockAssets(appBridge);
     const customIcon = blockAssets?.[ICON_ASSET_ID]?.[0];
 
+    const designSettingsStyleTag =
+        document.getElementById('theme-design-settings') ?? document.getElementById('design-settings')!;
+
     useEffect(() => {
         if (!blockSettings.iconSwitch && customIcon && isEditing) {
             deleteAssetIdsFromKey(ICON_ASSET_ID, [customIcon.id]);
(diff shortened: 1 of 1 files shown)


fulopdaniel:
Task linked: [CU-86989mrb0 Callout blocks need a page refresh to display the newly selected color scheme](https://app.clickup.com/t/86989mrb0)

packages/callout-block/src/CalloutBlock.tsx
ragi96, line 1: window.d.ts file can now be deleted

packages/callout-block/src/CalloutBlock.tsx
```diff
@@ -43,12 +46,16 @@ export const CalloutBlock = ({ appBridge }: BlockProps): ReactElement => {
...
 
-        window.emitter.on('HubAppearanceUpdated', updateStyles);
-
         updateStyles();
 
+        const styleChangeObserver = new MutationObserver(() => updateStyles());
+
+        if (designSettingsStyleTag) {
+            styleChangeObserver.observe(designSettingsStyleTag, { childList: true });
+        }
```
ragi96, line 55:
```suggestion
        if (!designSettingsStyleTag) {
            return;
        }
        const styleChangeObserver = new MutationObserver(() => updateStyles());
        styleChangeObserver.observe(designSettingsStyleTag, { childList: true, subtree: true });
```

oliverschwendener: Done ✅

ragi96:
Close in favor of https://github.com/Frontify/guideline-blocks/pull/1034

PR #1034, opened 2025-03-13
feat(Callout): support new color schemes

ragi96:
Instead of listening for the HubAppearanceUpdated event we can just observe the style tag that contains all CSS variables, to update the styles of the callout block. This makes sure that the callout block is compatible with the new design settings, which don't emit the HubAppearanceUpdated event anymore when changed.

CU-86989mrb0


fulopdaniel:
Task linked: [CU-86989mrb0 Callout blocks need a page refresh to display the newly selected color scheme](https://app.clickup.com/t/86989mrb0)

fulopdaniel (approved):
awesome

PR #1043, opened 2025-03-25
fix: add a fallback value for default color object instead of null when clearing input

peter-tudosa:
Ticket: CU-TASK-14020


ragi96 (approved):
LGTM

PR #1061, opened 2025-04-03
fix(CompareSliderBlock): add focus ring to handle

fulopdaniel:
bit of a workaround because the original button that is focused is inside the 3rd party package and not customisable.


fulopdaniel:
Task linked: [TASK-12526 Block: Compare Slider: focus not visible on slider button](https://app.clickup.com/t/2523021/TASK-12526)

peter-tudosa (approved):

PR #1063, opened 2025-04-03
fix(ImageBlock): add missing screen reader announcement for download

peter-tudosa:


fulopdaniel:
Task linked: [TASK-13162 Block: Image: download must be announced to screen reader via aria-live attribute](https://app.clickup.com/t/2523021/TASK-13162)

PR #1064, opened 2025-04-03
fix(QuoteBlock): add semantic elements for quote block

peter-tudosa:


fulopdaniel:
Task linked: [TASK-12939 Block: Quote: quote is not being announced as such to screen reader](https://app.clickup.com/t/2523021/TASK-12939)

PR #1066, opened 2025-04-04
fix(shared): add aria label to CSS value display

fulopdaniel:


fulopdaniel:
Task linked: [TASK-12932 Block: Gradient: input element is missing an aria-label](https://app.clickup.com/t/2523021/TASK-12932)

peter-tudosa (approved):

PR #1073, opened 2025-04-09
fix(DoDont): assets are stored definetly at the wrong place

ragi96:
Asset ids shouldn't be stored in the settings.....

Currently it leads to not working copy: TASK-14223

Goal here is that the block supports the bad structure (assetId within Item) and the new one storing the asset within as document block asset with the item id as identifier. 

Newly created should only use the good structure. Replacement should use the new structure. 

If this works as expected we will run a data migration and then we are able to remove the legacy part of the block (imageId on the item).

Currently known issue:
- [x] replace of correct asset will not delete old

Test Block Id on Product: `cm9b1p7a80001j1uqkciykz7d`


packages/dos-donts-block/src/DoDontItem.tsx
```diff
@@ -107,12 +107,15 @@ export const DoDontItem = memo((props: DoDontItemProps) => {
...
+                const asset = result[0];
+                const imageAlt = alt ?? asset.title ?? asset.fileName ?? '';
                 setLocalAltText(imageAlt);
+                if (deleteAssetIdsFromKey && linkedImage) {
+                    deleteAssetIdsFromKey(id, [linkedImage.id]);
+                }
                 if (addAssetIdsToKey) {
-                    addAssetIdsToKey(IMAGES_ASSET_KEY, [imageId]).then(() => {
-                        onChangeItem(id, { imageId, alt: imageAlt });
+                    addAssetIdsToKey(id, [asset.id]).then(() => {
```
fulopdaniel, line 117: Can we use `updateAssetIdsFromKey` instead of deleting and adding?

ragi96: Not fully sure I need to try out

ragi96: Replaced

fulopdaniel (approved):

PR #1083, opened 2025-04-14
fix(ImageBlock): Image title autofill

peter-tudosa:
* image title is automatically populated with the filename

### DIFF
--- a/packages/image-block/src/ImageBlock.tsx
+++ b/packages/image-block/src/ImageBlock.tsx
@@ -74,14 +74,12 @@ export const ImageBlock = withAttachmentsProvider(({ appBridge }: BlockProps) =>
         if (isFirstImageUpload) {
             const defaultImageName = newImage?.title ?? newImage?.fileName ?? '';
 
-            setBlockSettings({ altText: defaultImageName });
+            settings.altText = defaultImageName;
             setLocalAltText(defaultImageName);
 
             const hasManuallyEditedName = hasRichTextValue(name);
             if (!hasManuallyEditedName) {
                 settings.name = convertToRteValue(TextStyles.imageTitle, defaultImageName, 'center');
-                await setBlockSettings({ name: settings.name });
-                setTitleKey(generateRandomId());
             }
         }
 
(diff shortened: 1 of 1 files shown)


fulopdaniel:
Task linked: [CU-8698qefw3 🐞 Image block: Title no longer inherits the file name automatically](https://app.clickup.com/t/8698qefw3)

packages/image-block/src/ImageBlock.tsx
```diff
@@ -80,6 +80,7 @@ export const ImageBlock = withAttachmentsProvider(({ appBridge }: BlockProps) =>
             const hasManuallyEditedName = hasRichTextValue(name);
             if (!hasManuallyEditedName) {
                 settings.name = convertToRteValue(TextStyles.imageTitle, defaultImageName, 'center');
+                await setBlockSettings({ name: settings.name });
```
fulopdaniel, line 83: How come this fixes the issue? We call setBlockSettings a bit later in this function anyway.

peter-tudosa: This was a pretty "fun" one to debug. It turns out that there's a lot of memoization stuff going on in the RTE in Fondue, so once you are rendering the thing you can't really update the value anymore unless you unmount and mount again or you change the id (title key in our case here).

What was happening before this fix is that we were updating the key before updating the actual value for the RTE because `setBlockSettings` is async.

The fix makes sure that we are waiting for the `setBlockSettings` to finish and only after that we update the title key so that the RTE triggers a re-render.

fulopdaniel: A condition like `if (settings.name) { setTitleKey(generateRandomId() }` should take care of that, no? I also noticed that there is an extra setBlockSettings inside that branch as well regarding alt text, we can remove that as well and just have one in the end. This reduces api requests from 3 to 1 on first image upload (at the cost of 1 extra if).

```
    const updateImage = async (newImage: Asset) => {
        const settings: Partial<Settings> = {};
        const isFirstImageUpload = !image;

        if (isFirstImageUpload) {
            const defaultImageName = newImage?.title ?? newImage?.fileName ?? '';
            settings.altText = defaultImageName;
            setLocalAltText(defaultImageName);

            const hasManuallyEditedName = hasRichTextValue(name);
            if (!hasManuallyEditedName) {
                settings.name = convertToRteValue(TextStyles.imageTitle, defaultImageName, 'center');
            }
        }

        if (newImage.backgroundColor) {
            settings.backgroundColor = newImage.backgroundColor;
            settings.hasBackground = true;
        }

        if (Object.keys(settings).length > 0) {
            await setBlockSettings(settings);
        }
        await updateAssetIdsFromKey(IMAGE_ID, [newImage.id]);

        if (settings.name) {
            setTitleKey(generateRandomId());
        }
        setIsLoading(false);
    };
```

fulopdaniel (approved):

PR #1089, opened 2025-04-22
chore: bump pnpm and pnpm/action-setup

ragi96:
as title states

And also cypress and vite


package.json
```diff
@@ -53,5 +53,5 @@
             "@udecode/plate-emoji@31.4.0": "patches/@udecode__plate-emoji@31.4.0.patch"
         }
     },
-    "packageManager": "pnpm@9.15.9"
+    "packageManager": "pnpm@10.9.0"
```
SamuelAlev, line 56: with pnpm 10 you need `pnpm approve-builds` and allow the packages, if Cypress still fails you can ping me so I have a look.

ragi96: Yep, `pnpm rebuild` and then `pnpm approve-builds`, did the trick. Thanks for the hint.

PR #1099, opened 2025-05-19
fix: adjusted the alpha level on the helper text

benfrontify:
The title and description have already been fixed, so I only had to adjust the alpha level with Tailwind

### DIFF
--- a/packages/asset-kit-block/src/components/AssetCount.tsx
+++ b/packages/asset-kit-block/src/components/AssetCount.tsx
@@ -16,5 +16,9 @@ export const AssetCount = ({ count, color }: AssetCountProps): ReactElement => {
         );
     }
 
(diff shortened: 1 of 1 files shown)


packages/asset-kit-block/src/components/AssetCount.tsx
```diff
@@ -16,5 +16,5 @@ export const AssetCount = ({ count, color }: AssetCountProps): ReactElement => {
         );
     }
 
-    return <div className="tw-text-black-50">Add assets to make them available</div>;
+    return <div className="tw-text-black-50 tw-opacity-70">Add assets to make them available</div>;
```
fulopdaniel, line 19: You also need to change the color from black-50, to the body color defined in design settings.

To do so you can use

```
style={{
    color: BlockStyles[TextStyles.p].color,
}}
```

You can import these from 
`@frontify/guideline-blocks-settings`

PR #1100, opened 2025-05-21
fix(DoDont): unregister asset properly after item deletion

ragi96:
remove old asset handling and fix item deletion: TASK-14483

### DIFF
--- a/packages/dos-donts-block/src/DosDontsBlock.tsx
+++ b/packages/dos-donts-block/src/DosDontsBlock.tsx
@@ -39,7 +39,7 @@ import {
     generateRandomId,
 } from '@frontify/fondue';
 import { AssetsContext, AssetsProvider } from './AssetsProvider';
-import { CONTAINER_SMALL_LIMIT, DONT_ICON_ASSET_KEY, DO_ICON_ASSET_KEY, IMAGES_ASSET_KEY } from './const';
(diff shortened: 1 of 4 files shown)


packages/dos-donts-block/src/DosDontsBlock.tsx
```diff
@@ -274,19 +273,25 @@ export const DosDontsBlock: FC<BlockProps> = ({ appBridge }) => {
...
+
             setLocalItems((prevItems) => {
-                const itemToRemove = prevItems.find((item) => item.id === itemId);
-                const newItems = prevItems.filter((item) => item.id !== itemId);
-                setBlockSettings({ items: newItems });
-                if (itemToRemove?.imageId && deleteAssetIdsFromKey) {
-                    deleteAssetIdsFromKey(itemToRemove.id, [itemToRemove?.imageId]);
-                }
-                return newItems;
+                updatedItems = prevItems.filter((item) => item.id !== itemId);
```
fulopdaniel, line 279: As far as I know, the callback provided to state setters are asynchronous natures, so it is not guaranteed that updatedItems will be populated properly before `setBlockSettings` is called.

What about something like

```
const updatedItems = localItems.filter(item => item.id !== itemId)
setLocalItems(updatedItems)
setBlockSettings({items: updatedItems})
... bla bla bla
```

Also adding the `localItems` to the dependency array.

ragi96: I reverted that refactoring to keep the same logic as before just with the new call

PR #1103, opened 2025-05-30
fix: hide AssetKitBlock in view mode if empty

benfrontify:
Hide AssetKitBlock in view mode when the title and description are missing

### DIFF
--- a/packages/asset-kit-block/src/AssetKitBlock.tsx
+++ b/packages/asset-kit-block/src/AssetKitBlock.tsx
@@ -46,12 +46,11 @@ export const AssetKitBlock = ({ appBridge }: BlockProps): ReactElement | null =>
 
     const hasTitle = title && hasRichTextValue(title);
     const hasDescription = description && hasRichTextValue(description);
-    const shouldDisplayComponent = isEditing || hasTitle || hasDescription;
 
     const currentAssets = blockAssets[ASSET_SETTINGS_ID] ?? [];
+    const hasAssets = currentAssets.length > 0;
+    const shouldDisplayComponent = isEditing || hasTitle || hasDescription || hasAssets;
     const { generateBulkDownload, status, downloadUrl } = useAssetBulkDownload(appBridge);
-
-    const shouldHideComponent = currentAssets.length === 0 && !isEditing;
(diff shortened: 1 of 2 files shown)


fulopdaniel:
What we want is to hide the block when there are no assets available.

For this you can check the `currentAssets` array length.

Also, in this scenario I think it makes sense to just do an early return to avoid wrapping the whole component in a condition.

So the final solution could look something like this:


```
const shouldHideComponent = currentAssets.length === 0 && !isEditing;

if (!shouldDisplayComponent) {
        return null;
}

return (
        <div className="asset-kit-block">
                ...
        </div>
);
```

Also you need to adjust the tests. You can run `pnpm test:components:browser` on the root level of the repository to open cypress. You probably need to tweak the tests to make sure that the test runs in edit mode.

something like:
`withAppBridgeBlockStubs(AssetKitBlock, { editorState: true })`

packages/asset-kit-block/src/AssetKitBlock.tsx
```diff
@@ -44,7 +44,12 @@ export const AssetKitBlock = ({ appBridge }: BlockProps): ReactElement => {
         buttonText,
     } = blockSettings;
 
+    const hasTitle = title && hasRichTextValue(title);
+    const hasDescription = description && hasRichTextValue(description);
+
     const currentAssets = blockAssets[ASSET_SETTINGS_ID] ?? [];
+    const hasAssets = currentAssets.length > 0;
+    const shouldDisplayComponent = isEditing || hasTitle || hasDescription || hasAssets;
```
fulopdaniel, line 52: `hasTitle || hasDescription` should not be part of this condition, as far as I know if there are no assets uploaded, we don't want to display the asset-kit block.

fulopdaniel: Well, tough one.

Let's keep it. 👍 

Anyways, then you need to fix the lint errors in `AssetKitBlock.spec.ct.tsx`, you need to use single quotes instead of double. You can use the same approach I showed you on web-app.

fulopdaniel (approved):

PR #1137, opened 2025-08-07
fix(ImageBlock): fallback on true when assetViewerEnabled is undefined

hochreutenerl:
CU-869a23gw6

Currently, we have image block which do not have a value in the `assetViewerEnabled` setting.

So far the block fell back to not enabling the asset viewer. 
As in the past though the asset viewer was active by default and the default value of the setting is `true`, we decided to fall back to `true`.


fulopdaniel:
Task linked: [CU-869a23gw6 Asset viewer values](https://app.clickup.com/t/869a23gw6)

PR #1160, opened 2025-09-08
fix(ResponsiveImage): prevent layout shift 

peter-tudosa:
Prevent layout shift by making sure the image takes up space before loading

### DIFF
--- a/packages/image-block/src/ImageBlock.spec.ct.tsx
+++ b/packages/image-block/src/ImageBlock.spec.ct.tsx
@@ -62,6 +62,7 @@ describe('Image Block', () => {
     });
 
     it('should render the image responsively rounded to the nearest hundred pixel', () => {
+        window.devicePixelRatio = 1;
         cy.viewport(240, 800);
         const ImageBlockWithStubs = getImageBlockWithContainer({
             blockAssets: {


packages/shared/src/components/ResponsiveImage/ResponsiveImage.tsx
```diff
@@ -44,14 +45,23 @@ export const ResponsiveImage = ({
     const sourceOptimised = `${sourceWithWidth}${conversionParams}`;
 
     const dimensions = image.width && image.height ? { width: imageWidth, height: imageHeight } : {};
+    const handleImageLoad = useCallback(() => {
+        setImageLoaded(true);
+    }, [setImageLoaded]);
+
+    const stylesToApply = useMemo(() => {
+        return !imageLoaded ? { ...style, aspectRatio: `${imageWidth} / ${imageHeight}` } : style;
```
fulopdaniel, line 53: How does this work when `aspectRatio` is already set, e.g. in the Image Block when a ratio is set?

Maybe it's better to do it like this:

```suggestion
        return !imageLoaded ? { aspectRatio: `${imageWidth} / ${imageHeight}`, ...style  } : style;
```

So that the aspectRatio can be overriden if it's set.

packages/shared/src/components/ResponsiveImage/ResponsiveImage.tsx
```diff
@@ -44,14 +45,23 @@ export const ResponsiveImage = ({
...
+    }, [imageHeight, imageLoaded, imageWidth, style]);
 
     return (
         <img
             data-test-id={testId}
-            className={joinClassNames(['tw-flex tw-w-full', className])}
+            className={joinClassNames(['tw-flex tw-bg-box-neutral tw-w-full', className])}
             loading="lazy"
+            decoding="async"
```
fulopdaniel, line 61: I am not familiar with this attribute, could you explain why this is needed?

packages/shared/src/components/ResponsiveImage/ResponsiveImage.tsx
peter-tudosa: So the whole problem with the layout shift is happening because by default we have the aspect ratio set to 'auto'. Somehow when it's set to that it doesn't pre-allocate space in the layout and when it's finally loaded only then it takes the space that it's needed.

The strategy would be to apply an aspect ration based on the imageWidth and imageHeight that we have until the image is loaded, so that it allocates the proper space for the image and after it's loaded we take it out and let the aspect ratio that it's set by the user settings to take effect.

fulopdaniel (approved):

PR #1161, opened 2025-09-09
fix: default background alphalevel

benfrontify:
The alpha level should be at 0 when it's the default background

### DIFF
--- a/packages/template-block/src/settings.ts
+++ b/packages/template-block/src/settings.ts
@@ -30,7 +30,7 @@ import {
 } from './types';
 import { TEMPLATE_BLOCK_SETTING_ID } from './constants';
 
-export const DEFAULT_BACKGROUND_COLOR = { red: 255, green: 255, blue: 255, alpha: 1 };
+export const DEFAULT_BACKGROUND_COLOR = { red: 255, green: 255, blue: 255, alpha: 0 };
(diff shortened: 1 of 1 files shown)


ragi96:
@fulopdaniel can't this change existing Codesandbox blocks? as it's also used as fallback?

fulopdaniel:
> @fulopdaniel can't this change existing template blocks? as it's also used as fallback?

We tested that, and it did not change existing blocks. Reason being, on mount the default settings are saved to the db => alpha level 1 is actually saved on db level on existing blocks => defaults don't kick in.

fulopdaniel:
> > > @fulopdaniel can't this change existing template blocks? as it's also used as fallback?
> > 
> > 
> > We tested that, and it did not change existing blocks. Reason being, on mount the default settings are saved to the db => alpha level 1 is actually saved on db level on existing blocks => defaults don't kick in.
> 
> Isn't it used as fallback (didn't check code) I mean it's exported from the settings file

Yep, it's used as fallback, but ideally it never kicks in.

1. Block is mounted for the first time.
2. Default settings defined in settings.ts are saved to db
3. On subsequent loads, the settings from db are used.

This means that older blocks will still have `alpha=1` set as their background color, while new blocks will have `alpha=1`.

The default fallback in code only kicks in if for some reason the #2 step didn't work.

PR #1165, opened 2025-09-12
refactor(CompareSlider): improve image calculation

ragi96:
move tests from `cypress` to `vitest`
changed handling of the initial height
changed image handling to use `ResponsiveImage`
changed hight approach => switch to `aspectRatio`
added missing test cases

Still shitty but better

### DIFF
--- a/packages/compare-slider-block/src/CompareSliderBlock.tsx
+++ b/packages/compare-slider-block/src/CompareSliderBlock.tsx
@@ -33,7 +33,7 @@ import {
     UploadView,
 } from './components';
 import { BlockProps, THEME_PREFIX, radiusStyleMap, toRgbaString } from '@frontify/guideline-blocks-settings';
-import { StyleProvider } from '@frontify/guideline-blocks-shared';
+import { ResponsiveImage, StyleProvider, useImageContainer } from '@frontify/guideline-blocks-shared';
 
 import {
     DEFAULT_BACKGROUND_COLOR,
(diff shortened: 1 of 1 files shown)


packages/compare-slider-block/src/CompareSliderBlock.tsx
```diff
@@ -214,44 +189,24 @@ export const CompareSliderBlock = ({ appBridge }: BlockProps) => {
...
-        if (!firstAsset || !secondAsset || !currentSliderWidth) {
-            return 0;
+    const getContainerAspectRatio = useMemo((): string => {
+        if (!firstAsset || !secondAsset) {
+            return '16/9';
         }
+        const firstAspectRatio = firstAsset[0].width / firstAsset[0].height;
+        const secondAspectRatio = secondAsset[0].width / secondAsset[0].height;
+        const minAspectRatio = Math.min(firstAspectRatio, secondAspectRatio);
+        return `${minAspectRatio}/1`;
```
fulopdaniel, line 199: Previously the compare slider was as tall as the shortest image. Now that logic is reversed, and the taller image takes precedence.

sooo

```
const maxAspectRatio = Math.max(firstAspectRatio, secondAspectRatio);
return String(maxAspectRatio)
```

fulopdaniel: https://www.loom.com/share/f136eb1fb0a8468e9e44486b3a4efa1a

ragi96: Yep see that, but also taking the max leads to even more changes, but yes I see we are not there yet

PR #1305, opened 2026-04-15
refactor: treat columnNumber as a string

hochreutenerl:
CU-869ctj5ec

`columnNumber` is used in a segmented control, which means it needs to be treated as a string.


fulopdaniel:
Task linked: [CU-869ctj5ec Text (new)](https://app.clickup.com/t/869ctj5ec)

fulopdaniel (approved):

silviojaeger (approved):

PR #1309, opened 2026-04-16
refactor: improve typing in image block settings

hochreutenerl:
CU-869ctj5hg

This is an internal refactoring (no UI / behaviour changes) that improves the handling of null values and removes some unused code / properties.

All settings that are marked as nullable, are actually not defined when creating a new image block.

### DIFF
--- a/packages/image-block/src/types.ts
+++ b/packages/image-block/src/types.ts
@@ -8,7 +8,6 @@ export type Link = { link: { link: string }; openInNewTab: boolean };
 export type Settings = {
     hasLink?: boolean;
     linkObject?: Link;
-    image?: string;
     name?: string;
     altText?: string;
     description?: string;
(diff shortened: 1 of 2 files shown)


fulopdaniel:
Task linked: [CU-869ctj5hg Image](https://app.clickup.com/t/869ctj5hg)

packages/image-block/src/types.ts
```diff
@@ -123,13 +122,6 @@ export enum Autosizing {
     Fit = 'fit',
     Fill = 'fill',
 }
-
-export const autosizingValues: Record<Autosizing, string> = {
```
silviojaeger, line 127: Are we sure this is not used somewhere outside of this repo? Im not super familiar with this

packages/image-block/src/types.ts
```diff
@@ -22,20 +21,20 @@ export type Settings = {
     hasBorder: boolean;
     hasCustomPadding: boolean;
     paddingChoice: Padding;
-    paddingCustom: string;
+    paddingCustom?: string;
```
silviojaeger, line 24: This means the settings are optional (not nullable like you described).
So paddingCustom can be a string or undefined. `paddingCustom=null` would break. Is this the expected behaviour?

packages/image-block/src/types.ts
hochreutenerl: A search in the Frontify organization shows no other [results](https://github.com/search?q=org%3AFrontify+autosizingValues&type=code)

packages/image-block/src/types.ts
hochreutenerl: It did not really break, but just display no padding. An alternative could be the following:
```
const getPadding = (blockSettings: Settings): string =>
    blockSettings.hasCustomPadding ? (blockSettings.paddingCustom ?? '0px') : paddingValues[blockSettings.paddingChoice];
```

silviojaeger (approved):

PR #1313, opened 2026-04-20
fix: append px to divider thickness

hochreutenerl:
When entering a number (without px) into the divider block, "px" is not appended automatically, causing this [error](https://sentryapp.appsupport.frontify.dev/organizations/frontify/issues/981608/?project=48)

### DIFF
--- a/packages/divider-block/src/settings.ts
+++ b/packages/divider-block/src/settings.ts
@@ -224,6 +224,7 @@ export const settings = defineSettings({
                     placeholder: 'e.g. 3px',
                     clearable: false,
                     rules: [numericalOrPixelRule, minimumNumericalOrPixelOrAutoRule(1)],
+                    onChange: (bundle) => appendUnit(bundle, THICKNESS_ID),
                 },
                 {
                     id: 'color',


packages/divider-block/src/settings.ts
```diff
@@ -224,6 +224,7 @@ export const settings = defineSettings({
                     placeholder: 'e.g. 3px',
                     clearable: false,
                     rules: [numericalOrPixelRule, minimumNumericalOrPixelOrAutoRule(1)],
+                    onChange: (bundle) => appendUnit(bundle, THICKNESS_ID),
```
silviojaeger, line 227: maybe? to make intention more clear?


```suggestion
                    onChange: (bundle) => appendUnit(bundle, THICKNESS_ID, 'px'),
```

silviojaeger (approved):

packages/divider-block/src/settings.ts
hochreutenerl: We overmitted that in all the other cases cases too. I will keep it as it is atm.

PR #1320, opened 2026-04-28
refactor: add image block schema

hochreutenerl:
CU-869ctj5hg

This adds the `settingsSchema` for the image block.
As some settings are now always given, we can mark them as non-nullable in the code too.
`fondueRte` fields are still missing `frontify-api-write: true` as we don't have a validator for that yet.


fulopdaniel:
Task linked: [CU-869ctj5hg Image](https://app.clickup.com/t/869ctj5hg)

silviojaeger (commented):
🤞

silviojaeger (approved):

PR #1336, opened 2026-05-11
refactor: migrate image block tests to vitest

ragi96:
Adds a skill for migrating from cypress to vitest, also migrates the image block


fulopdaniel (approved):
Cool!

PR #1337, opened 2026-05-11
test(Shared): move to vitest

ragi96:
as title states (created by the skill), review properly and then let's adjust the skill with findings


peter-tudosa (approved):

ragi96:
verify-manifest fails because of changes on the app-server

PR #1340, opened 2026-05-13
feat(FigmaBlock): check for asset status to be finished before usage

peter-tudosa:
## The problem we're trying to solve:

When the asset status is not FINISHED and we're trying to use the `previewUrl` we're putting a lot of strain on the worker that deals with the assets.

## Solution
We're going to check if the `asset.status` is `FINISHED` before using the `previewUrl`.

Technical details:
* create a custom react hook `useAssetStatusPolling` that will be used in `FigmaImagePreview` to communicate if the asset is ready
* the hook does the checking of the status and polls for updates if the status is not FINISHED
* we show a loading circle if the asset is not ready

### DIFF
--- a/packages/figma-block/src/FigmaBlock.tsx
+++ b/packages/figma-block/src/FigmaBlock.tsx
@@ -167,6 +167,9 @@ export const FigmaBlock = ({ appBridge }: BlockProps): ReactElement => {
                                     title={asset.title}
                                     url={asset.previewUrl}
                                     assetExternalUrl={safeExternalUrl}
+                                    assetId={asset.id}
+                                    assetStatus={asset.status}
+                                    appBridge={appBridge}
                                     hasLimitedOptions={hasLimitedOptions}
                                     height={isCustomHeight ? heightValue : heights[heightChoice]}
                                     hasBorder={hasBorder}
--- a/packages/figma-block/src/components/FigmaImagePreview.tsx
+++ b/packages/figma-block/src/components/FigmaImagePreview.tsx
@@ -29,7 +29,7 @@ export const FigmaImagePreview = ({
     return (
         <div data-test-id="figma-image-preview" className="tw-flex tw-flex-col tw-justify-center">
             {isReady ? (
-                <ImageStage {...imageStageProps} url={imageStageProps.url} />
+                <ImageStage {...imageStageProps} />
             ) : (
                 <div className="tw-flex tw-items-center tw-justify-center tw-w-full tw-py-8">
                     <LoadingCircle />
--- a/packages/figma-block/src/components/FigmaImagePreview.tsx
+++ b/packages/figma-block/src/components/FigmaImagePreview.tsx
@@ -31,7 +31,7 @@ export const FigmaImagePreview = ({
             {isReady ? (
                 <ImageStage {...imageStageProps} />
             ) : (
-                <div 
+                <div
                     className="tw-flex tw-items-center tw-justify-center tw-w-full"
                     style={{ height: imageStageProps.height }}
                 >
--- a/packages/figma-block/src/components/FigmaImagePreview.tsx
+++ b/packages/figma-block/src/components/FigmaImagePreview.tsx
@@ -31,7 +31,10 @@ export const FigmaImagePreview = ({
             {isReady ? (
                 <ImageStage {...imageStageProps} />
             ) : (
-                <div className="tw-flex tw-items-center tw-justify-center tw-w-full tw-py-8">
+                <div 
+                    className="tw-flex tw-items-center tw-justify-center tw-w-full"
+                    style={{ height: imageStageProps.height }}
+                >
                     <LoadingCircle />
                 </div>
             )}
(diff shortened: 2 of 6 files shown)


fulopdaniel:
Task linked: [CU-869d82vwy FigmaSync: media delivery issue](https://app.clickup.com/t/869d82vwy)

packages/figma-block/src/FigmaBlock.tsx
```diff
@@ -167,6 +167,9 @@ export const FigmaBlock = ({ appBridge }: BlockProps): ReactElement => {
                                     title={asset.title}
                                     url={asset.previewUrl}
                                     assetExternalUrl={safeExternalUrl}
+                                    assetId={asset.id}
+                                    assetStatus={asset.status}
+                                    appBridge={appBridge}
```
ragi96, line 172:
```suggestion
                                    asset={asset}
                                    assetExternalUrl={safeExternalUrl}
                                    appBridge={appBridge}
```

Would love to simplify that one, wdyt?

packages/figma-block/src/hooks/useAssetStatusPolling.ts
```diff
@@ -0,0 +1,64 @@
...
+            }
+
+            if (attempts >= MAX_POLL_ATTEMPTS) {
+                console.error('[useAssetStatusPolling] Max poll attempts reached for asset', assetId);
+                clearInterval(intervalId);
+            }
+        };
+
+        intervalId = setInterval(() => {
+            poll().catch(console.error);
```
ragi96:
```suggestion
            poll()
```

already caught

packages/figma-block/src/hooks/useAssetStatusPolling.ts
```diff
@@ -0,0 +1,64 @@
...
+): UseAssetStatusPollingReturn => {
+    const [pollingDetectedFinished, setPollingDetectedFinished] = useState(false);
+    const isReady = isFinished(initialStatus) || pollingDetectedFinished;
+
+    useEffect(() => {
+        if (isFinished(initialStatus)) {
+            return;
+        }
+
+        let intervalId: number;
```
ragi96:
```suggestion
        let intervalId: ReturnType<typeof setInterval>;
```

packages/figma-block/src/hooks/useAssetStatusPolling.ts
```diff
@@ -0,0 +1,64 @@
...
+
+const ASSET_STATUS_FINISHED = 'FINISHED';
+const POLLING_INTERVAL_MS = 3000;
+const MAX_POLL_ATTEMPTS = 20;
+
+type UseAssetStatusPollingReturn = {
+    isReady: boolean;
+};
+
+const isFinished = (status: string | undefined) => status && status === ASSET_STATUS_FINISHED;
```
ragi96:
```suggestion
const isFinished = (status: string | undefined): boolean => status === ASSET_STATUS_FINISHED;
```
NIT

packages/figma-block/src/components/FigmaImagePreview.tsx
```diff
@@ -1,29 +1,53 @@
...
+    ...imageStageProps
+}: FigmaImagePreviewProps) => {
+    const { isReady } = useAssetStatusPolling(appBridge, assetId, assetStatus);
+
+    return (
+        <div data-test-id="figma-image-preview" className="tw-flex tw-flex-col tw-justify-center">
+            {isReady ? (
+                <ImageStage {...imageStageProps} />
+            ) : (
+                <div className="tw-flex tw-items-center tw-justify-center tw-w-full tw-py-8">
```
ragi96:
```suggestion
                <div 
                    className="tw-flex tw-items-center tw-justify-center tw-w-full"
                    style={{ height: imageStageProps.height }}
                >
```

to avoid layout shifts

packages/figma-block/src/FigmaBlock.tsx
peter-tudosa: I'd rather send just what is necessary since it's not that much.

packages/figma-block/src/hooks/useAssetStatusPolling.ts
peter-tudosa: Yea, the problem is the eslint that bugs me about this fn being async.

peter-tudosa: I can probably await it though.

packages/figma-block/src/FigmaBlock.tsx
ragi96: Fine for me

packages/figma-block/src/hooks/useAssetStatusPolling.ts
peter-tudosa: Went back and forth with this one. To me it's just cognitive load not offering anything real...

ragi96: Well basically it's the correct type, would go with it, even it's not offering something great, yet. But really no hard feelings from my end

ragi96 (approved):

PR #1345, opened 2026-05-19
fix: remove hardcoded link color in Figma block

benfrontify:


fulopdaniel:
Task linked: [CU-869d9cc5e change link color in image mode](https://app.clickup.com/t/869d9cc5e)

PR #1349, opened 2026-05-21
Remove: index.ts in helpers

benfrontify:


fulopdaniel:
Task linked: [CU-869dd0c99 Fix imports so they don't go through the index](https://app.clickup.com/t/869dd0c99)

PR #1355, opened 2026-05-27
feat: add machine translatable field

ragi96:


hochreutenerl (commented):
The selection is correct, I will approve as soon as the PR is out of draft

packages/image-block/manifest.json
```diff
@@ -63,12 +64,14 @@
                         "type": ["object", "null"],
                         "required": ["link"],
                         "frontify-translatable": true,
+                        "frontify-machine-translatable": false,
```
silviojaeger: This can be removed, also the translatable key. 
They should not live on a object, only on their properties. 
The documentation will explain this✌️

ragi96: So `verifyManifest` will also give you these hints?

silviojaeger: Not "yet"😉
I mean they are not invalid and do not break the schema, they are just obsolete.
We can add some functionality to the validation if needed👍

packages/image-block/manifest.json
```diff
@@ -64,13 +65,13 @@
                     "link": {
                         "type": ["object", "null"],
                         "required": ["link"],
-                        "frontify-translatable": true,
                         "frontify-api-read": true,
```
hochreutenerl: The read and write can be removed also, as they propagate too

hochreutenerl (approved):

PR #1357, opened 2026-05-28
fix(figma-block): make ImageStage fullscreen a proper modal (portal, dialog role, escape, scroll lock)

benfrontify:


fulopdaniel:
Task linked: [CU-869df1cpw ImageStage.tsx fullscreen overlay is not a modal](https://app.clickup.com/t/869df1cpw)

packages/figma-block/src/FullscreenOverlay.tsx
```diff
@@ -0,0 +1,65 @@
...
+            ref={overlayRef}
+            role="dialog"
+            aria-modal="true"
+            tabIndex={-1}
+            className="tw-fixed tw-top-0 tw-left-0 tw-w-full tw-h-full tw-z-[200] tw-outline-none"
+        >
+            {children}
+        </div>,
+        document.body
+    );
```
ragi96: Remount-on-toggle regression risk. The two branches return structurally different trees (Fragment in one, host <div> inside a portal in the other), so React unmounts/remounts the children subtree on every fullscreen toggle. That means stageRef/containerRef/imageRef re-point to new DOM nodes, the  re-mounts, and the ResizeObserver set up in useImageStage ([] deps) keeps observing the now-detached old stage node, resize-driven recalculations will stop working inside fullscreen.

Please verify zoom + pan still work after entering fullscreen, not just before. To keep the DOM stable, only toggle the portal target and the dialog attrs:

```
    const content = (
        <div
            ref={overlayRef}
            {...(isFullScreen && {
                role: 'dialog',
                'aria-modal': true,
                tabIndex: -1,
                className: 'tw-fixed tw-top-0 tw-left-0 tw-w-full tw-h-full tw-z-[200] tw-outline-none',
            })}
        >
            {children}
        </div>
    );

    return isFullScreen ? createPortal(content, document.body) : content;
```

packages/figma-block/src/FullscreenOverlay.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+    if (!isFullScreen) {
+        return <>{children}</>;
+    }
+
+    return createPortal(
+        <div
+            ref={overlayRef}
+            role="dialog"
+            aria-modal="true"
```
ragi96: `aria-modal="true"` is a lie without a focus trap. Tabbing from inside the overlay walks straight into the underlying page (still in the DOM behind the portal). Use fondue modal. Without it, the a11y attributes promise something the implementation doesn't deliver, and this is exactly the gap the PR is trying to close.

packages/figma-block/src/FullscreenOverlay.tsx
```diff
@@ -0,0 +1,65 @@
...
+        };
+    }, [isFullScreen]);
+
+    useEffect(() => {
+        if (!isFullScreen) {
+            return;
+        }
+
+        overlayRef.current?.focus();
+    }, [isFullScreen]);
```
ragi96: Focus is not restored on close. Capture the previously-focused element on open and restore it in the cleanup, otherwise screen-reader / keyboard users lose their place when leaving fullscreen.

```
    useEffect(() => {
        if (!isFullScreen) {
            return;
        }

        const previouslyFocused = document.activeElement as HTMLElement | null;
        overlayRef.current?.focus();
        return () => previouslyFocused?.focus();
    }, [isFullScreen]);
```

packages/figma-block/src/FullscreenOverlay.tsx
```diff
@@ -0,0 +1,65 @@
...
+    useEffect(() => {
+        if (!isFullScreen) {
+            return;
+        }
+
+        document.body.style.overflow = 'hidden';
+        return () => {
+            document.body.style.overflow = '';
+        };
+    }, [isFullScreen]);
```
ragi96: Don't do this, as it's manipulating something outside of the block, imo if you use the right components in fondue (need to check which one), does this

packages/figma-block/src/FullscreenOverlay.tsx
```diff
@@ -0,0 +1,65 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
```
ragi96: Should be in the folder `components`

ragi96 (commented):
PR title nits

Typo: "mondial" → "modal"
Title describes the file mechanic ("new file") rather than the user-visible change
Suggested: fix(figma-block): make ImageStage fullscreen a proper modal (portal, dialog role, escape, scroll lock)

packages/figma-block/src/components/FullscreenOverlay.tsx
```diff
@@ -0,0 +1,70 @@
...
+            }
+        };
+
+        document.addEventListener('keydown', handleKeyDown);
+
+        return () => {
+            document.removeEventListener('keydown', handleKeyDown);
+            previouslyFocused?.focus();
+        };
+    }, [isFullScreen, onClose]);
```
ragi96, line 71: Now the scroll lock is removed, but it's important, can you maybe use a component of fondue which portals and scroll locks, not sure if there is one

packages/figma-block/src/components/FullscreenOverlay.tsx
```diff
@@ -0,0 +1,70 @@
...
+                const last = focusable[focusable.length - 1];
+
+                if (e.shiftKey && document.activeElement === first) {
+                    e.preventDefault();
+                    last?.focus();
+                } else if (!e.shiftKey && document.activeElement === last) {
+                    e.preventDefault();
+                    first?.focus();
+                }
+            }
```
ragi96, line 58: Initial Tab from the wrapper escapes the trap. Right after open, overlayRef.current?.focus() runs so document.activeElement is the tabIndex={-1} wrapper. Pressing Shift+Tab then doesn't satisfy === first or === last, so the handler does nothing and the browser walks focus backward in document order — straight out of the modal. Repro: open fullscreen, immediately press Shift+Tab without clicking anything else.

Suggested guard: if the active element is the overlay itself or sits outside the overlay subtree, force focus to first (or last for Shift+Tab):


```suggestion
            if (e.key === 'Tab' && overlayRef.current) {
                const focusable = Array.from(overlayRef.current.querySelectorAll<HTMLElement>(FOCUSABLE_SELECTORS));
                if (focusable.length === 0) {
                    e.preventDefault();
                    overlayRef.current.focus();
                    return;
                }
                const first = focusable[0];
                const last = focusable[focusable.length - 1];
                const active = document.activeElement as HTMLElement | null;
                const isInsideOverlay = active && overlayRef.current.contains(active);

                if (e.shiftKey && (!isInsideOverlay || active === first)) {
                    e.preventDefault();
                    last.focus();
                } else if (!e.shiftKey && (!isInsideOverlay || active === last)) {
                    e.preventDefault();
                    first.focus();
                }
            }

```

packages/figma-block/src/components/FullscreenOverlay.tsx
```diff
@@ -0,0 +1,70 @@
...
+        };
+
+        document.addEventListener('keydown', handleKeyDown);
+
+        return () => {
+            document.removeEventListener('keydown', handleKeyDown);
+            previouslyFocused?.focus();
```
ragi96:
```suggestion
            if (previouslyFocused && document.contains(previouslyFocused)) {
                previouslyFocused.focus();
            }
```

packages/figma-block/src/components/FullscreenOverlay.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+    const content = (
+        <div
+            ref={overlayRef}
+            {...(isFullScreen && {
+                role: 'dialog',
+                'aria-modal': true,
+                tabIndex: -1,
+                className: 'tw-fixed tw-top-0 tw-left-0 tw-w-full tw-h-full tw-z-[200] tw-outline-none',
+            })}
```
ragi96:
```suggestion
            ref={overlayRef}
            role={isFullScreen ? 'dialog' : undefined}
            aria-modal={isFullScreen || undefined}
            tabIndex={isFullScreen ? -1 : undefined}
            className={isFullScreen ? 'tw-fixed tw-top-0 tw-left-0 tw-w-full tw-h-full tw-z-[200] tw-outline-none' : undefined}
```

Nicer to read

packages/figma-block/src/components/FullscreenOverlay.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { useEffect, useRef } from 'react';
+import { createPortal } from 'react-dom';
+
+const FOCUSABLE_SELECTORS =
+    'a[href], button:not([disabled]), input:not([disabled]), select:not([disabled]), textarea:not([disabled]), [tabindex]:not([tabindex="-1"])';
+
+type FullscreenOverlayProps = {
+    isFullScreen: boolean;
+    onClose: () => void;
+    children: React.ReactNode;
```
ragi96:
```suggestion
import { type ReactNode, useEffect, useRef } from 'react';
import { createPortal } from 'react-dom';

const FOCUSABLE_SELECTORS =
    'a[href], button:not([disabled]), input:not([disabled]), select:not([disabled]), textarea:not([disabled]), [tabindex]:not([tabindex="-1"])';

type FullscreenOverlayProps = {
    isFullScreen: boolean;
    onClose: () => void;
    children: ReactNode;
```

packages/figma-block/src/components/FullscreenOverlay.tsx
benfrontify: did not find one in fondue

ragi96: Then we need to do it manually, to have the scroll lock :)

packages/figma-block/src/components/FullscreenOverlay.tsx
ragi96: Not resolved

ragi96 (commented):
The 2 open comments still need work.

packages/figma-block/src/components/FullscreenOverlay.tsx
```diff
@@ -0,0 +1,88 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
```
ragi96, line 1: Almost there but can you add some tests for this component

- getByRole('dialog') with aria-modal="true" only when isFullScreen
- Pressing Escape calls onClose
- Children render in place (no portal) when !isFullScreen
- document.body.style.overflow is 'hidden' while open, restored to its prior value on close
- Focus is restored to the previously-focused element on close
Trap-behavior tests are harder under jsdom but the structural pieces above are easy wins.

ragi96 (commented):
Finish Line

ragi96 (approved):

PR #1362, opened 2026-06-04
refactor: add fondue compnent to figma live modal

benfrontify:


ragi96 (commented):
Nice one but now you can remove, the scroll-lock logic as the `Dialog` Component will already handle that

ragi96 (approved):

PR #1366, opened 2026-06-08
fix: chanched the map comon styles and add tests

benfrontify:


packages/figma-block/src/helpers/mapCommonStyle.spec.ts
```diff
@@ -0,0 +1,31 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { describe, expect, it } from 'vitest';
+
+import { getHeightOfBlock } from './mapCommonStyle';
+
+describe('getHeightOfBlock', () => {
```
ragi96, line 7: add a test with input `400, true`, return should be `400px`

ragi96: Sorry I was wrong `400, true` needs to be `200px` and `400, false` needs to be `400px`

packages/figma-block/src/helpers/mapCommonStyle.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+    it('should return the px height unchanged on desktop', () => {
+        expect(getHeightOfBlock('400px', false)).toBe('400px');
+    });
+
+    it('should return half of the px height on mobile', () => {
+        expect(getHeightOfBlock('400px', true)).toBe('200px');
+    });
+
+    it('should return a unitless height as px unchanged on mobile', () => {
+        expect(getHeightOfBlock('400', true)).toBe('400px');
```
ragi96:
```suggestion
        expect(getHeightOfBlock('400', true)).toBe('200px');
```

Should be 200 as stated above

ragi96 (commented):
See co

packages/figma-block/src/helpers/mapCommonStyle.ts
```diff
@@ -9,6 +9,15 @@ export const getBorderOfBlock = (hasBorder: boolean, borderStyle: string, border
...
+
+    if (!Number.isNaN(heightWithoutUnit) && !heightInSettings.match(/[a-z%]+$/i)) {
+        return `${heightWithoutUnit * (isMobile ? MOBILE_HEIGHT_MODIFIER : 1)}px`;
+    }
+
+    if (!heightInSettings.endsWith('px')) {
+        return heightInSettings;
+    }
+
     return `${heightWithoutUnit * (isMobile ? MOBILE_HEIGHT_MODIFIER : 1)}px`;
```
ragi96: Blocks (1) and (3) have byte-identical returns, so the function can collapse to a single guard + one return. `Number()` (instead of `parseFloat` + regex) handles the pass-through check cleanly — `Number('50vh')` is `NaN`, so non-px values still pass through. Passes all 7 tests unchanged.

```suggestion
    if (!heightInSettings.endsWith('px') && Number.isNaN(Number(heightInSettings))) {
        return heightInSettings;
    }

    const heightWithoutUnit = Number.parseFloat(heightInSettings);
    return `${heightWithoutUnit * (isMobile ? MOBILE_HEIGHT_MODIFIER : 1)}px`;
```

ragi96 (approved):

PR #1367, opened 2026-06-11
fix: Mouse listerner leaks

benfrontify:


packages/figma-block/src/components/ImageStage.ts
```diff
@@ -12,18 +12,23 @@ export class ImageStage {
...
-        this.boundaries = this.imageStage.getBoundingClientRect();
-        document.addEventListener('mousemove', this.checkIfMouseIsInside.bind(this));
+        this.boundaries = this.getBoundaries();
+        document.addEventListener('mousemove', this.checkIfMouseIsInside);
     }
 
-    private checkIfMouseIsInside(event: MouseEvent) {
+    private readonly checkIfMouseIsInside = (event: MouseEvent) => {
         const currentMousePosition = MouseProperties.getCurrentPosition(event);
+        this.boundaries = this.getBoundaries();
```
ragi96: `getBoundingClientRect()` now runs on every document-wide `mousemove`. It forces synchronous layout when layout is dirty, and `mousemove` fires at very high frequency — with several blocks on a page this multiplies. The old cached value was stale after scroll, but a cheaper fix is to keep the cache and invalidate it on `scroll`/`resize`, or recompute at most once per frame:

```ts
private boundariesDirty = false;

private readonly invalidateBoundaries = () => {
    this.boundariesDirty = true;
};
// constructor: window.addEventListener('scroll', this.invalidateBoundaries, true);
//              window.addEventListener('resize', this.invalidateBoundaries);
// destroy():   remove both

private readonly checkIfMouseIsInside = (event: MouseEvent) => {
    if (this.boundariesDirty) {
        this.boundaries = this.getBoundaries();
        this.boundariesDirty = false;
    }
    // ...
};
```

If per-event recomputation is intentional, please add a comment explaining why.

packages/figma-block/src/hooks/useImageStage.ts
```diff
@@ -49,6 +62,8 @@ export const useImageStage = ({ height, hasLimitedOptions, isMobile }: UseImageS
 
         const resizeObserver = new ResizeObserver(() => {
             if (imageRef.current && containerRef.current && imageStageRef.current) {
+                containerOperatorRef.current?.destroy?.();
```
ragi96: `destroy` is declared on the abstract `ContainerOperator` base, so it always exists — the `?.()` after `destroy` is unnecessary and would hide a type error if the method were ever removed.

```suggestion
                containerOperatorRef.current?.destroy();
```

packages/figma-block/src/hooks/useImageStage.ts
```diff
@@ -64,7 +79,11 @@ export const useImageStage = ({ height, hasLimitedOptions, isMobile }: UseImageS
         });
         resizeObserver.observe(stageRef.current);
 
-        return () => resizeObserver.disconnect();
+        return () => {
+            resizeObserver.disconnect();
+            containerOperatorRef.current?.destroy?.();
```
ragi96: Same as above, `destroy` always exists on the base class:

```suggestion
            containerOperatorRef.current?.destroy();
```

packages/figma-block/src/components/ImageStage.ts
```diff
@@ -36,7 +41,11 @@ export class ImageStage {
 
     public alterHeight(height: string) {
         this.imageStage.style.height = height;
-        this.boundaries = this.imageStage.getBoundingClientRect();
+        this.boundaries = this.getBoundaries();
+    }
+
+    public destroy(): void {
```
ragi96: No tests were added for the new cleanup behavior, even though `ImageStage.spec.tsx` and `useImageStage.spec.ts` exist. The leak fix is very testable — spy on `document.addEventListener`/`removeEventListener` and assert `destroy()` removes the exact handler that was added:

```ts
it('removes the mousemove listener on destroy', () => {
    const addSpy = vi.spyOn(document, 'addEventListener');
    const removeSpy = vi.spyOn(document, 'removeEventListener');

    const stage = new ImageStage(document.createElement('div'));
    const handler = addSpy.mock.calls.find(([type]) => type === 'mousemove')?.[1];

    stage.destroy();

    expect(removeSpy).toHaveBeenCalledWith('mousemove', handler);
});
```

ragi96 (changes requested):
**`destroy()` is never overridden in `VectorContainerOperator`, so the operator cleanup added in this PR is currently a no-op — and that's where the remaining listener leaks are.**

`VectorContainerOperator.ts:25-26` adds `mouseover`/`mousedown` listeners to `imageContainer.node` with inline `.bind(this)` — the same un-removable pattern this PR fixes in `ImageStage`. The ResizeObserver callback creates a new operator on every resize against the same persistent DOM node, so listeners accumulate and every stale operator instance is retained by its closures.

Additionally, the drag `mousemove` listener on `document` is only removed by `mouseup` on the container node (`VectorContainerOperator.ts:71-75`). Releasing the mouse outside the container, or replacing the operator mid-drag, leaves a document-level `mousemove` handler attached forever.

Suggested override in `VectorContainerOperator.ts`:

```ts
private readonly onMouseOverListener = () => this.onMouseOver();
private readonly onMouseDownListener = (event: MouseEvent) => this.onMouseDown(event);

// in the constructor, replace lines 25-26 with:
imageContainer.node.addEventListener('mouseover', this.onMouseOverListener);
imageContainer.node.addEventListener('mousedown', this.onMouseDownListener);

public override destroy(): void {
    this.imageContainer.node.removeEventListener('mouseover', this.onMouseOverListener);
    this.imageContainer.node.removeEventListener('mousedown', this.onMouseDownListener);
    this.imageContainer.node.removeEventListener('mouseup', this.mouseUpListener);
    document.removeEventListener('mousemove', this.mouseMoveListener);
}
```

ragi96 (commented):
The updated revision looks much better, the `VectorContainerOperator.destroy()` override, the stored listener references, and the tests address the main leak concerns. Two changes I'd like to see before merging, plus a follow-up ticket:

**1. Fix the coordinate mismatch — the inside-check is still wrong when the page is scrolled, and the new scroll/resize invalidation can't fix it.** `MouseProperties.getCurrentPosition` returns `pageX/pageY` (document-relative, includes scroll), but `getBoundingClientRect()` is viewport-relative. With the page scrolled down by `S`, `pageY = clientY + S`, so `pageY < boundaries.bottom` fails even with a freshly recomputed rect. The new spec doesn't catch this because it dispatches events at scroll position 0, where `pageX === clientX`.

**2. Remove the document-level listener instead of managing it.** `isMouseInsideImageStage` has exactly one consumer — `VectorContainerOperator.onMouseMove`, which only runs during a drag (attached on mousedown, removed on mouseup). The dirty flag, the scroll/resize listeners, the three removals in `destroy()`, and their tests all exist to keep an always-on document-wide `mousemove` listener alive for a cursor tweak during drags. Computing the value on demand deletes all of that machinery and fixes point 1 in the same move:

```ts
// ImageStage — drop isMouseInsideImageStage, the mousemove/scroll/resize listeners, and the dirty flag
public isPointInside({ x, y }: Point): boolean {
    const rect = this.imageStage.getBoundingClientRect();
    return x > rect.left && x < rect.right && y > rect.top && y < rect.bottom;
}
```

```ts
// VectorContainerOperator.onMouseMove
if (this.imageStage.isPointInside({ x: event.clientX, y: event.clientY })) {
    this.imageContainer.changeMouseCursor(Cursor.GRABBING);
}
```

`getBoundingClientRect()` then runs only during drags (cheap), `ImageStage.destroy()` shrinks or disappears entirely, and there's nothing always-on left to leak — which also means the three new `ImageStage` specs reduce to one. The hook-side cleanup and the `VectorContainerOperator.destroy()` work stay as they are; that part is solid.

(If there's a consumer of `isMouseInsideImageStage` I'm missing, let me know and we can revisit.)

**Follow-up (separate ticket, not this PR):** migrate the drag in `VectorContainerOperator` to Pointer Events + `setPointerCapture`. That removes its remaining document-level drag listener, fixes the "mouseup outside the container leaves the listener attached" edge case, and adds touch/pen support for free — after it, nothing in this feature attaches document/window listeners at all.

packages/figma-block/src/components/ImageStage.ts
```diff
@@ -36,9 +26,17 @@ export class ImageStage {
 
     public alterHeight(height: string) {
         this.imageStage.style.height = height;
-        this.boundaries = this.imageStage.getBoundingClientRect();
+        this.boundaries = this.getBoundaries();
+    }
+
+    public isPointInside({ x, y }: { x: number; y: number }): boolean {
```
ragi96: Minor: there's already a `Point` type used elsewhere in this feature (e.g. `VectorContainerOperator`), could reuse it here instead of the inline shape for consistency.

```suggestion
    public isPointInside({ x, y }: Point): boolean {
```
(add `import { type Point } from '../types';`)

packages/figma-block/src/components/ImageStage.ts
```diff
@@ -36,9 +26,17 @@ export class ImageStage {
...
+        const boundaries = this.getBoundaries();
+
+        return x > boundaries.left && x < boundaries.right && y > boundaries.top && y < boundaries.bottom;
     }
 
+    public destroy(): void {}
```
ragi96:
```suggestion
```

`destroy()` is now an empty no-op, kept only so the hook can call it uniformly. Harmless, but it's effectively dead code on `ImageStage` now dropping it and only calling `destroy()` on the operator.

packages/figma-block/src/components/ContainerOperator/ContainerOperator.ts
```diff
@@ -16,6 +16,8 @@ export abstract class ContainerOperator {
         protected imageElement: ImageElement
     ) {}
 
+    public destroy(): void {}
```
ragi96: Make this `public abstract destroy(): void;` rather than an empty default. An empty base no-op is exactly what let the `VectorContainerOperator` override get missed in the first revision of this PR, forcing each subclass to declare it (an explicit one-liner in `BitmapContainerOperator`, which has no listeners) makes the decision visible.

packages/figma-block/src/components/ContainerOperator/VectorContainerOperator.ts
```diff
@@ -12,8 +12,11 @@ import { MAGNIFICATION_PERCENTAGE_DEFAULT } from './constants';
 export class VectorContainerOperator extends ContainerOperator {
```
ragi96, line 12: The core of this fix: `VectorContainerOperator.destroy()` actually removing its listeners, isn't covered by a test. Worth a spec that spies on `removeEventListener` (document `mousemove` + the container-node listeners) and asserts `destroy()` removes them, to guard against this regressing.

packages/figma-block/src/components/ContainerOperator/BitmapContainerOperator.ts
```diff
@@ -7,6 +7,9 @@ import { type ImageStage } from '../ImageStage';
 import { ContainerOperator } from './ContainerOperator';
 
 export class BitmapContainerOperator extends ContainerOperator {
+    public destroy(): void {
+        throw new Error('Method not implemented.');
+    }
```
ragi96:
```suggestion
public destroy(): void {}
```

This should be a noop. With this error it will break the block

packages/figma-block/src/components/ContainerOperator/VectorContainerOperator.ts
```diff
@@ -48,7 +51,7 @@ export class VectorContainerOperator extends ContainerOperator {
     }
 
     private onMouseMove(event: MouseEvent) {
```
ragi96, line 53: mouseup is registered on imageContainer.node while mousemove is on document, so releasing the button outside the container never ends the drag.

ragi96 (approved):

PR #1370, opened 2026-06-15
refactor(doAndDont): extracted Color to a new file

benfrontify:
refactor doAndDontBlock:
currently exracting all the cod that dosent have to be in doAndDontBlock.tsx in this case color


packages/dos-donts-block/src/helpers/Color.ts
```diff
@@ -0,0 +1,15 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
```
ragi96, line 1: Naming of the file and placement is wrong

ragi96, line 1: file should be `.ts`, tests are missing

ragi96 (approved):
LGTM

PR #1376, opened 2026-06-17
feat(doAndDontBlock): extracted Grid to a new file GCM-197

benfrontify:


packages/dos-donts-block/src/helper/Grid.ts
```diff
@@ -0,0 +1,16 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
```
ragi96, line 1: make a `helper` folder and call it `GridHelper`, also it can be `.ts` not `.tsx`

ragi96 (commented):
Also there is an Issue with the component test

packages/dos-donts-block/src/helper/Grid.ts
```diff
@@ -0,0 +1,19 @@
...
```
ragi96, line 1: Maybe would even just call it `Grid`, as everything within the `dos-donts-block` is related to the DosDonts stuff

ragi96, line 1: Also add tests for this helper

ragi96 (approved):

PR #1383, opened 2026-06-30
feat(dos-donts): add assets file and tests

benfrontify:
add assets file and tests


packages/dos-donts-block/src/DoDontItem.tsx
```diff
@@ -1,7 +1,6 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
```
ragi96, line 2: The ref is already typed `useRef<DosDontsAssetsRef>(null)` and `DosDontsAssetsRef.openUpload` is non-optional, so this cast is redundant and actively harmful, it re-types the ref to a weaker shape where the method is optional. If openUpload is ever renamed/removed on the ref, TypeScript won't flag this call and the menu action silently no-ops at runtime. Call the typed ref directly.

Also for the other not only openUpload

ragi96 (changes requested):

packages/dos-donts-block/src/DoDontItem.tsx
```diff
@@ -385,7 +309,7 @@ export const SortableDoDontItem = memo((props: SortableDoDontItemProps) => {
         editing ? { ...attributes, ...listeners } : {}
     );
     const transformStyle = {
-        transform: transform ? `translate(${transform.x}px, ${transform.y}px)` : '',
+        transform: transform ? `translate(${transform.x}px)` : '',
```
ragi96: Items render in a CSS grid (`tw-grid`, 1–4 columns) sorted with dnd-kit's `rectSortingStrategy`, so both X and Y offsets are needed to slide siblings into their new slots. In a 1-column/narrow layout all motion is vertical, so siblings get a zero-effect horizontal translate and don't move at all, in a multi-row grid, items that change rows animate horizontally only and jump/overlap. This looks accidental — please restore `transform.y`.



```suggestion
        transform: transform ? `translate(${transform.x}px, ${transform.y}px)` : '',
```

packages/dos-donts-block/src/DosDontsAssets.tsx
```diff
@@ -0,0 +1,209 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
```
ragi96: Get rid of the eslint disablement, or have a good reason to add them :)

packages/dos-donts-block/src/DosDontsAssets.tsx
```diff
@@ -0,0 +1,209 @@
...
+
+        const onOpenAssetChooser = () => {
+            openAssetChooser(
+                // eslint-disable-next-line @typescript-eslint/no-misused-promises
+                async (result: Asset[]) => {
+                    setIsUploadLoading(true);
+
+                    const asset = result[0];
+                    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+                    const imageAlt = alt ?? asset.alternativeText ?? '';
```
ragi96: Both compute `imageAlt = alt ?? asset.alternativeText ?? ''`, call `setLocalAltText`, `await updateAssetIdsFromKey(id, [asset.id])`, `onChangeItem(id, { alt })`, and `setIsUploadLoading(false)`, but the ordering of `setIsUploadLoading(false)` vs `onChangeItem` differs between them. Extract a single `applyAsset(asset)` helper so future fixes land in one place.

packages/dos-donts-block/src/DosDontsAssets.tsx
```diff
@@ -0,0 +1,209 @@
...
+                uploadFile(selectedFiles);
+            }
+            // eslint-disable-next-line @eslint-react/exhaustive-deps
+        }, [selectedFiles]);
+
+        useEffect(() => {
+            if (doneAll) {
+                // eslint-disable-next-line @typescript-eslint/no-floating-promises
+                (async (uploadResults) => {
+                    const asset = uploadResults?.[0];
```
ragi96:
```suggestion
                    const asset = uploadResults?.[0];
                    if (!asset) {
                        return;
                    }
```

packages/dos-donts-block/src/DosDontsAssets.spec.tsx
```diff
@@ -0,0 +1,389 @@
...
+        const chooserCallback = mocks.openAssetChooser.mock.calls[0]?.[0] as (
+            result: Array<{ id: number; alternativeText?: string }>
+        ) => Promise<void>;
+
+        await act(async () => {
+            await chooserCallback([
+                {
+                    id: 123,
+                    alternativeText: 'Asset alt text',
+                },
```
ragi96: The `EditAltTextFlyout` mock renders a Save button bound to `onSave`, but no test clicks it to assert `onChangeItem(id, { alt: localAltText })` fires. Chooser/upload alt handling is well covered; the flyout's own persistence path could break with the suite still green. Please add a test that opens the menu, clicks Save, and asserts the `onChangeItem` call.

packages/dos-donts-block/src/DosDontsAssets.tsx
```diff
@@ -0,0 +1,217 @@
...
+                return;
+            }
+            const asset = uploadResults?.[0];
+
+            if (!asset) {
+                return;
+            }
+
+            saveAsset(asset).catch(() => undefined);
+        }, [doneAll, saveAsset, uploadResults]);
```
ragi96:
```suggestion
        useEffect(() => {
            if (!doneAll) {
                return;
            }
            const asset = uploadResults?.[0];

            if (!asset) {
                return;
            }

            saveAsset(asset).catch(() => undefined);
            // eslint-disable-next-line @eslint-react/exhaustive-deps
        }, [doneAll, uploadResults]);
```

`saveAsset` is a `useCallback` keyed on [`alt`, `id`, `onChangeItem`, `updateAssetIdsFromKey`], so its identity changes whenever alt changes. After an upload finishes, `useAssetUpload's` doneAll/uploadResults stay set (that's exactly why the original excluded the functions from its deps). So this effect now re-runs on any alt change while doneAll is still true.

packages/dos-donts-block/src/DosDontsAssets.tsx
```diff
@@ -0,0 +1,217 @@
...
+            }),
+            [onOpenAssetChooser, onUploadClick]
+        );
+
+        useEffect(() => {
+            if (selectedFiles) {
+                setIsUploadLoading(true);
+                uploadFile(selectedFiles);
+            }
+        }, [selectedFiles, uploadFile]);
```
ragi96:
```suggestion
        // eslint-disable-next-line @eslint-react/exhaustive-deps
        }, [selectedFiles]);
```

If `useAssetUpload's` `uploadFile` is not referentially stable across renders, this effect re-runs while `selectedFiles` is still populated and re-uploads the same file. It's the same avoidable regression

ragi96 (commented):
This 2 are important ones, If you don't get why we need to surpress it here, please reach out to me

packages/dos-donts-block/src/DosDontsAssets.tsx
```diff
@@ -0,0 +1,217 @@
...
+
+                if (updateAssetIdsFromKey) {
+                    await updateAssetIdsFromKey(id, [asset.id]);
+                    onChangeItem(id, { alt: imageAlt });
+                }
+
+                setIsUploadLoading(false);
+            },
+            [alt, id, onChangeItem, updateAssetIdsFromKey]
+        );
```
ragi96:
```suggestion
const saveAsset = useCallback(
    async (asset: Asset) => {
        const imageAlt = getImageAltText(alt, asset);
        setLocalAltText(imageAlt);
        try {
            if (updateAssetIdsFromKey) {
                await updateAssetIdsFromKey(id, [asset.id]);
                onChangeItem(id, { alt: imageAlt });
            }
        } finally {
            setIsUploadLoading(false);
        }
    },
    [alt, id, onChangeItem, updateAssetIdsFromKey]
);
```

packages/dos-donts-block/src/DosDontsAssets.tsx
```diff
@@ -0,0 +1,217 @@
...
+            hasBackground,
+            hasRadius,
+            radiusChoice,
+            borderColor,
+            borderStyle,
+            borderWidth,
+            hasBorder,
+            radiusValue,
+            dontColor,
+        },
```
ragi96: Prop drilling

packages/dos-donts-block/src/DosDontsAssets.tsx
```diff
@@ -0,0 +1,217 @@
...
+                    extensions: FileExtensionSets.Images,
+                }
+            );
+        }, [closeAssetChooser, openAssetChooser, saveAsset]);
+
+        const onUploadClick = useCallback(() => {
+            openFileDialog();
+        }, [openFileDialog]);
+
+        useImperativeHandle(
```
ragi96: Works but `useImperativeHandle` isn't an often used idiom, so I would prefer a different solution can also be done as a follow up

benfrontify: ok i will make a new ticket for this one

packages/dos-donts-block/src/DosDontsAssets.tsx
```diff
@@ -0,0 +1,193 @@
...
+                showAltTextMenu={showAltTextMenu}
+                setLocalAltText={setLocalAltText}
+                defaultAltText={alt}
+                onSave={() => onChangeItem(id, { alt: localAltText })}
+                localAltText={localAltText}
+            />
+
+            {mode === BlockMode.TEXT_AND_IMAGE && (
+                <ImageComponent
+                    {...props}
```
ragi96: Still drills the props, but we can change it later, but never pass pass props like this it's an anti pattern imo

packages/dos-donts-block/src/DoDontItem.tsx
```diff
@@ -63,20 +61,13 @@ export const DoDontItem = memo((props: DoDontItemProps) => {
         setActivatorNodeRef,
         alt,
     } = props;
+
+    const titleRef = useRef<HTMLTextAreaElement>(null);
```
ragi96: `titleRef` is never assigned to any dom object, not sure if it's not used anymore or you are creating a bug here

packages/dos-donts-block/src/DosDontsAssets.tsx
ragi96: @benfrontify never pass props

benfrontify:
I’m not 100% sure if I resolved the drilling issue. Let me know how this attempt looks.

fulopdaniel:
> I’m not 100% sure if I resolved the drilling issue. Let me know how this attempt looks.

@benfrontify can you kindly explain to me the drilling issue you are trying to solve and your attempt at solving it?

By drilling issue I suspect it's about avoiding to send down a bunch of props through several components?

benfrontify:
> {...props} i used this and Raffi commented on this that I shouldn't do thrilling or pass them one by one

packages/dos-donts-block/src/DoDontItem.tsx
```diff
@@ -188,67 +132,77 @@ export const DoDontItem = memo((props: DoDontItemProps) => {
...
-                        backgroundColor={backgroundColor}
-                        hasBackground={hasBackground}
-                        hasRadius={hasRadius}
-                        radiusChoice={radiusChoice}
-                        border={hasBorder ? `${borderWidth} ${borderStyle} ${toRgbaString(borderColor)}` : ''}
-                        radiusValue={radiusValue}
-                        dontColor={dontColor}
-                    />
-                )}
+                <DosDontsAssets
```
fulopdaniel: I think something like this would already work:

`<DosDontsAssets ref={assetsRef} {...props} />`

Can you try this one? Then you can avoid manually writing all the props.

fulopdaniel: Only render <DosDontsAssets /> if the mode is `BlockMode.TEXT_AND_IMAGE`

packages/dos-donts-block/src/DosDontsAssets.tsx
```diff
@@ -0,0 +1,268 @@
...
+>;
+
+const getImageAltText = (alt: string | undefined, asset: Asset): string => {
+    return (
+        alt ??
+        (typeof asset.alternativeText === "string" ? asset.alternativeText : "")
+    );
+};
+
+const toRgbaString = (
```
fulopdaniel: Don't we already have this implemented somewhere? E.g. `@frontify/guideline-blocks-settings`

packages/dos-donts-block/src/DosDontsAssets.tsx
```diff
@@ -0,0 +1,268 @@
...
+    | "radiusChoice"
+    | "borderColor"
+    | "borderStyle"
+    | "borderWidth"
+    | "hasBorder"
+    | "radiusValue"
+    | "dontColor"
+>;
+
+const getImageAltText = (alt: string | undefined, asset: Asset): string => {
```
fulopdaniel: helper functions should typically be moved out to a new helper file, to keep the component thin.

packages/dos-donts-block/src/DosDontsAssets.tsx
```diff
@@ -0,0 +1,268 @@
...
+                extensions: FileExtensionSets.Images,
+            },
+        );
+    }, [closeAssetChooser, openAssetChooser, saveAsset]);
+
+    const onUploadClick = useCallback(() => {
+        openFileDialog();
+    }, [openFileDialog]);
+
+    useImperativeHandle(
```
fulopdaniel: please close the other PR and include that in this one as they will heavily conflict.

packages/dos-donts-block/src/DosDontsAssets.tsx
```diff
@@ -0,0 +1,268 @@
...
+        }
+        const asset = uploadResults?.[0];
+        if (!asset) {
+            return;
+        }
+        saveAsset(asset).catch(() => undefined);
+        // eslint-disable-next-line @eslint-react/exhaustive-deps
+    }, [doneAll, uploadResults]);
+
+    const imageProps: ImageComponentProps = {
```
fulopdaniel: If you will use my `{...props}` suggestion, this can be removed and the component is simplified to:

```
                <ImageComponent
                    onAssetChooseClick={onOpenAssetChooser}
                    onUploadClick={onUploadClick}
                    isUploadLoading={isUploadLoading}
                    isEditing={editing}
                    border={hasBorder ? `${borderWidth} ${borderStyle} ${toRgbaString(borderColor)}` : ''}
                    isDragging={Boolean(isDragging)}
                    {...props}
                />
```

packages/dos-donts-block/src/DosDontsAssets.tsx
```diff
@@ -0,0 +1,268 @@
...
+            <EditAltTextFlyout
+                setShowAltTextMenu={setShowAltTextMenu}
+                showAltTextMenu={showAltTextMenu}
+                setLocalAltText={setLocalAltText}
+                defaultAltText={alt}
+                onSave={() => onChangeItem(id, { alt: localAltText })}
+                localAltText={localAltText}
+            />
+
+            {mode === BlockMode.TEXT_AND_IMAGE && (
```
fulopdaniel: The condition can be removed and moved outside the component inside DoDontItem.tsx.

Don't render this component at all if we are not in this mode.

packages/dos-donts-block/src/DosDontsAssets.tsx
ragi96: I am personally not a big fan of passing `props`

fulopdaniel: Then what about something like this.

1. Provide the block settings via context
2. Create `useDoDontItem()` and `useDoDontAsset()` hooks and mount inside the respective component. These consume the block settings and return the necessary callbacks and values - things that right now we pass as props.

packages/dos-donts-block/src/hooks/useDoDontAssets.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+
+describe('useDoDontAssets', () => {
+    beforeEach(() => {
+        vi.clearAllMocks();
+        uploadState.selectedFiles = undefined;
+        uploadState.results = undefined;
+        uploadState.doneAll = false;
+    });
+
+    it('opens the file dialog', () => {
```
fulopdaniel: All test cases should start with the word "should"

packages/dos-donts-block/src/DoDontItem.tsx
```diff
@@ -193,62 +140,62 @@ export const DoDontItem = memo((props: DoDontItemProps) => {
...
-                        hasRadius={hasRadius}
-                        radiusChoice={radiusChoice}
-                        border={hasBorder ? `${borderWidth} ${borderStyle} ${toRgbaString(borderColor)}` : ''}
-                        radiusValue={radiusValue}
-                        dontColor={dontColor}
+                        hasStrikethrough={
+                            type === DoDontType.Dont && hasStrikethrough
+                        }
+                        border={
+                            hasBorder
```
fulopdaniel: `border` and `hasStrikethrough` has some inline calculations here. Move those calculations to `useDoDontAssets` so all the logic is there, and just return `border` and `hasStrikethrough` from that hook.

packages/dos-donts-block/src/DoDontItem.spec.tsx
```diff
@@ -0,0 +1,73 @@
...
+const renderDoDontItem = (props: Partial<DoDontItemProps> = {}) => render(<DoDontItem {...defaultProps} {...props} />);
+
+describe('DoDontItem', () => {
+    it('renders the assets in text and image mode', () => {
+        renderDoDontItem({ mode: BlockMode.TEXT_AND_IMAGE });
+
+        expect(screen.getByTestId(DOS_DONTS_ASSETS)).toBeInTheDocument();
+    });
+
+    it('does not render the assets in text mode', () => {
```
fulopdaniel: "should not" here as well wording.

fulopdaniel (approved):

PR #1390, opened 2026-07-13
fix(Template): avoid double calls on Template select

ragi96:
Follow up to https://github.com/Frontify/app-server/pull/32381

So the block doesn't call onTemplateSelect twice, so it's always only template per block.


fulopdaniel (approved):
I don't know the context, the solution works, so I'll approve this one, but is it possible to avoid emitting twice in the first place?

ragi96:
> I don't know the context, the solution works, so I'll approve this one, but is it possible to avoid emitting twice in the first place?

Tbh I searched for the root cause but didn't find it, we just saw it in the logs that it happened. 🙈

findmind (approved):
Change makes sense and looks clean. 👌 Have not built the block and tested it, but let's go!

What about updating? Do we need to roll out the update on customer instances as well, or is this not important enough? (iirc we have a script to auto-update a Frontify block on all instances for critical stuff)

PR #1392, opened 2026-07-14
feat(DoDontBlock): new style file

benfrontify:
extracted styles


packages/dos-donts-block/src/DoDontStyle.tsx
```diff
@@ -0,0 +1,98 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
```
ragi96: Do one hook per style that goes together, this is now doing way to much

ragi96 (changes requested):

benfrontify:
I am not 100% sure if every file should starts with use or if this is ok

ragi96:
> I am not 100% sure if every file should starts with use or if this is ok

Would create a folder `hooks` for the `useDoDontColorStyle` (also add tests for it)
The other 2 should be in the `helpers` folder

packages/dos-donts-block/src/helpers/getDoDontGridStyle.ts
```diff
@@ -0,0 +1,26 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { getGridClassName } from '../helpers/Grid';
```
ragi96:
```suggestion
import { getGridClassName } from './Grid';
```

ragi96 (approved):
LGTM

PR #1398, opened 2026-07-21
Gcm 335 make abetter solution then drilling props

benfrontify:
follow-up task from [Frontify/guideline-blocks#1383](https://linear.app/frontify/review/featdos-donts-add-assets-file-and-tests-616a1d8d5fcc)

### DIFF
--- a/packages/dos-donts-block/src/DoDontItem.tsx
+++ b/packages/dos-donts-block/src/DoDontItem.tsx
@@ -176,22 +176,12 @@ export const DoDontItem = memo((props: DoDontItemProps) => {
                                           {
                                               title: 'Replace with upload',
                                               icon: <IconArrowCircleUp size={20} />,
-                                              onClick: () => {
-                                                  const assets = assetsRef.current as {
-                                                      openUpload?: () => void;
-                                                  } | null;
-                                                  assets?.openUpload?.();
-                                              },
+                                              onClick: () => assetsRef.current?.openUpload(),
                                           },
                                           {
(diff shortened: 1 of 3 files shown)


peter-tudosa:
Not sure what the "brief" is for this task but...

It looks to me that we are just moving the problem around without really solving it while complicating things even more by passing references and using the `useImperativeHandle`.
Essentially, to me, it looks like we create even more  layers to prop drill. 🤷 

A funny thing that caught my eye is that we are passing along `updateAssetIdsFromKey` through many layers when this exists on a context provider and we could just reach out for it.

Anyways, what I would do is I would take a step back, and try to extract the logic from all this first into custom hook, try to trace out how data flows and see exactly how much prop drilling are we doing and if it's bad or not.

Prop drilling per say is not a bad thing in my opinion, sometimes you just need to pass stuff around.

If what bothers us is stuff like:
```js
<SomeComponent
  prop1={prop1}
  prop2={prop2}
  ....
  propN={propN}
```

we could create a variable before hand, use the shorthand way for declaring the props and pass the object around while spreading the props:

```js
const myFunProps = {prop1, prop2, ... propN}
...
<SomeComponent {...myFunProps} />
```

We could even organize them a bit if needed, I saw that some of them are styles that we pass around to reach Image component, we could create a variable for than gather in there are the props pertaining to image styles and decide if we want to send the object or spread it.

@ragi96  @benfrontify  Again I'm not sure what we're trying to solve here but these are my observations.

ragi96:
Yep see it the same way

PR #1399, opened 2026-07-21
chore: migrate to oxfmt and oxlint

peter-tudosa:
Ditching `prettier` and `eslint` for `oxfmt` and `oxlint`.

‼️ ~~Not 100% ready to merge due to:~~
- ~~`deploy` command fails at the moment because `frontify-cli deploy` runs underneath `npx eslint src` -- we need to address this in the cli package first~~

Depends on: https://github.com/Frontify/brand-sdk/pull/1628


fulopdaniel (approved):

packages/animation-curve-block/src/components/Card.tsx
```diff
@@ -190,7 +190,7 @@ export const Card = forwardRef<HTMLDivElement, CardProps>(
                                 isEditing={isEditing}
                             />
                         )}
-                        <div className={merge([hasBorder && 'tw-px-4', 'tw-text-small tw-test-primary'])}>
+                        <div className={merge([hasBorder && 'tw-px-4', 'tw-text-small'])}>
```
ragi96: You sure about dropping this one? Basically I assume it should have been `tw-text-primary` before 🤷

examples/asset-upload/oxlint.config.ts
```diff
@@ -0,0 +1,47 @@
...
+                        trailingEmptyLines: {
+                            minimum: 2,
+                        },
+                    },
+                ],
+                'tailwindcss/no-custom-classname': [
+                    'warn',
+                    {
+                        whitelist: ['^(?!tw-).*$'],
+                        callees: ['className', 'sv', 'class', 'merge'],
```
ragi96: Wondering if `joinClassNames` is missing here

ragi96 (commented):
Also should we switch from `eslint-disable` to `oxlint-disable`?

packages/animation-curve-block/src/components/Card.tsx
peter-tudosa: tw-test-primary is not defined anywhere, linter was complaining so I took it out cause it looked odd.

ragi96: Yeah I know fine to drop, but I assume who ever did it (me)? Intended to do `tw-text-primary` lmao

peter-tudosa: A lol you think it should be text-primary. I can put it back in the right form, so it was a typo in the beginning. 😆

ragi96: Well not 100% sure, but would be reasonable 🤷

examples/asset-upload/oxlint.config.ts
peter-tudosa: Yup I will add it everywhere.

ragi96 (approved):

PR #1413, opened 2026-08-04
feat(*): remove modal extensions from postcss scoping

fulopdaniel:
Instead of adding a bunch of modal related selectors, we now use ThemeProvider className. Adjusted all blocks, mostly just whitespace changes.

**Before:**

```
<div className="text-block">
    <StyleProvider>
        [...]
    </StyleProvider>
</div>
```

```
.text-block .myClass, body > [role='toolbar'] .myClass, body > [role="dialog"] .myClass { ... }
```

**After:**

```
<StyleProvider className="text-block">
    [...]
</StyleProvider>
```

```
.text-block .myClass { ... }
```

### DIFF
--- a/examples/asset-upload/block-scope.json
+++ b/examples/asset-upload/block-scope.json
@@ -1,0 +1,3 @@
+{
+    "scope": "example-asset-upload"
+}
--- a/examples/asset-upload/postcss.config.js
+++ b/examples/asset-upload/postcss.config.js
@@ -1,9 +1,11 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
+const blockScope = require('./block-scope.json');
+
 module.exports = {
     plugins: [
         require('tailwindcss'),
         require('autoprefixer'),
-        require('../../postcss/scope')({ scope: '.example-asset-upload' }),
+        require('../../postcss/scope')({ scope: blockScope.scope }),
     ],
 };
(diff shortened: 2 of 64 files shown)


peter-tudosa:
Looks pretty good, I have a few things I would improve.
I noticed we use the same token (class) in `postcss.config.js` and the main `*block.tsx` file where we pass it to the Styleguide component.

I would:
1. create a `block-scope.json` file, to keep the scope in there as a single source of truth
  - probably .json will be easier for everyone to read, also easy I think, if at one point we decide to make this a part of the manifest

2. in Styleguide I would pass a prop named `scope` (instead of classname) and make it required, I think we always want to enforce scoping, right?

fulopdaniel:
> Looks pretty good, I have a few things I would improve. I noticed we use the same token (class) in `postcss.config.js` and the main `*block.tsx` file where we pass it to the Styleguide component.
> 
> I would:
> 
> 1. create a `block-scope.json` file, to keep the scope in there as a single source of truth
> 
> * probably .json will be easier for everyone to read, also easy I think, if at one point we decide to make this a part of the manifest
> 
> 2. in Styleguide I would pass a prop named `scope` (instead of classname) and make it required, I think we always want to enforce scoping, right?

adjusted @peter-tudosa 👍

peter-tudosa (approved):

PR #1418, opened 2026-08-11
fix: cleanup of lint stuff

benfrontify:


benfrontify:
Whenever I save all the ' switches to " dk if thats a problem

packages/dos-donts-block/src/DoDontItem.tsx
```diff
@@ -208,7 +217,7 @@ export const DoDontItem = memo((props: DoDontItemProps) => {
                 {style === DoDontStyle.Underline && (
                     <hr
                         style={dividerStyles[type]}
-                        className="tw-w-full tw-my-3 tw-h-[3px] tw-border-none tw-rounded-medium tw-bg-black-40"
+                        className="tw-w-full tw-my-3 tw-h-[3px] tw-border-none tw-rounded-medium"
```
ragi96, line 220: Was this color never used?

benfrontify: No, so it has the color of the title, so it doesn't make a difference if it is here or not. It was confusing because I was searching for this element, so I deleted this part.

ragi96 (approved):

PR #1421, opened 2026-08-13
fix(CalloutBlock): compute accessible text colors at WCAG AA 4.5:1

peter-tudosa:
- Raise callout text contrast to WCAG AA 4.5:1
- Light: keep the accent when it already passes, otherwise mix it toward black/white
- Strong: pick white or black from contrast
- Tests assert AA with TinyColor instead of hardcoded mixer RGBs

### DIFF
--- a/packages/callout-block/package.json
+++ b/packages/callout-block/package.json
@@ -24,6 +24,7 @@
     },
     "devDependencies": {
         "@babel/core": "^7.29.0",
+        "@ctrl/tinycolor": "^4.2.0",
         "@frontify/frontify-cli": "^6.2.6",
         "@frontify/oxfmt-config": "^0.0.2",
         "@frontify/oxlint-config-react": "^0.0.2",
--- a/packages/callout-block/src/CalloutBlock.spec.ct.tsx
+++ b/packages/callout-block/src/CalloutBlock.spec.ct.tsx
@@ -1,11 +1,16 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
+import { readability } from '@ctrl/tinycolor';
 import { AssetDummy, withAppBridgeBlockStubs } from '@frontify/app-bridge';
 import { mount } from 'cypress/react';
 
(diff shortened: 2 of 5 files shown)


fulopdaniel:
would 3.5 be enough?

peter-tudosa:
> would 3.5 be enough?

I dunno, in here it says 4.5 for normal text: https://www.w3.org/WAI/standards-guidelines/act/rules/afw4f7/proposed/

The tools that I use to check the contrast also use this value, as well as `readability` from `tinycolor` which I know we don't use here per say but in other places we do so I figured let's stick with this value.

fulopdaniel:
> > would 3.5 be enough?
> 
> I dunno, in here it says 4.5 for normal text: https://www.w3.org/WAI/standards-guidelines/act/rules/afw4f7/proposed/
> 
> The tools that I use to check the contrast also use this value, as well as `readability` from `tinycolor` which I know we don't use here per say but in other places we do so I figured let's stick with this value.

Yeah I think that's the industry standard, but maybe we increase this value step by step to avoid suddenly changing a lot of callout colors.

cc @ragi96 what's your take on this?

ragi96:
> > > would 3.5 be enough?
> > 
> > 
> > I dunno, in here it says 4.5 for normal text: [w3.org/WAI/standards-guidelines/act/rules/afw4f7/proposed](https://www.w3.org/WAI/standards-guidelines/act/rules/afw4f7/proposed/)
> > The tools that I use to check the contrast also use this value, as well as `readability` from `tinycolor` which I know we don't use here per say but in other places we do so I figured let's stick with this value.
> 
> Yeah I think that's the industry standard, but maybe we increase this value step by step to avoid suddenly changing a lot of callout colors.
> 
> cc @ragi96 what's your take on this?

Let's go perfect by now, and let's see how customers responed.

@LidijaStoilova will be happy :D

fulopdaniel (approved):

PR #1443, opened 2026-09-01
feat(codeSnippet): replace cypress with vitest

benfrontify:


packages/code-snippet-block/src/CodeSnippetBlock.spec.tsx
```diff
@@ -0,0 +1,200 @@
...
+import userEvent from '@testing-library/user-event';
+import { type ReactNode } from 'react';
+import { describe, expect, it, vi } from 'vitest';
+
+import { CodeSnippetBlock } from './CodeSnippetBlock';
+
+vi.mock('./styles.css', () => ({}));
+vi.mock('@frontify/guideline-blocks-shared', () => ({
+    StyleProvider: ({ children }: { children: ReactNode }) => children,
+}));
```
ragi96:
```suggestion
vi.mock('./styles.css', () => ({}));
vi.mock('@frontify/guideline-blocks-shared', async () => {
    const actual = await vi.importActual<typeof GuidelineBlocksShared>('@frontify/guideline-blocks-shared');
    return {
        ...actual,
        StyleProvider: ({ children }: { children: ReactNode }) => children,
    };
})
```

packages/code-snippet-block/src/CodeSnippetBlock.spec.tsx
```diff
@@ -0,0 +1,200 @@
...
+    });
+
+    it('should copy the content using the copy button without a header', async () => {
+        const user = userEvent.setup();
+        const content = `const counter = function() {
+        let count = 0;
+        return function() {
+            return ++count;
+        }
+    };`;
```
ragi96:
```suggestion
        const content = `const counter = function() {
            let count = 0;
            return function() {
                return ++count;
            }
        };`;
```

Format properly

package.json
```diff
@@ -27,6 +27,7 @@
         "@4tw/cypress-drag-drop": "^2.3.1",
         "@babel/core": "^7.29.0",
         "@cypress/vite-dev-server": "^7.2.1",
+        "@frontify/guideline-blocks-settings": "4.0.0-alpha.1",
```
ragi96:
```suggestion
```

Is this really needed?

ragi96: You need to remove it and the run pnpm i

ragi96 (approved):

PR #1445, opened 2026-09-01
feat(code-snippet-block): component folder and copy button component

benfrontify:
Tests will be added after the PR is merged to cover the Cypress-to-Vitest test-file fixes.


packages/code-snippet-block/src/CodeSnippetBlock.tsx
```diff
@@ -24,12 +24,14 @@ export const CodeSnippetBlock: FC<BlockProps> = ({ appBridge }) => {
...
     const isEditing = useEditorState(appBridge);
     const [contentValue] = useState(blockSettings.content);
-    const [pendingLanguage, setPendingLanguage] = useState<Language>();
-    const selectedLanguage = pendingLanguage ?? blockSettings.language ?? 'plain';
+    const [selectedLanguage, setSelectedLanguage] = useState(blockSettings.language ?? 'plain');
     const extensions = useCodeMirrorExtensions(selectedLanguage);
-    const [isCopied, setIsCopied] = useState(false);
-    const [isCopyTooltipOpen, setIsCopyTooltipOpen] = useState(false);
     const labelId = useMemo(() => `${appBridge.context('blockId').get()}-header`, [appBridge]);
+
```
ragi96:
```suggestion
```

ragi96:
Resolve the conflicts first :)

ragi96 (commented):
Fix the CI

packages/code-snippet-block/src/components/CopyButton.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+export const CopyButton: FC<CopyButtonProps> = ({ content, className, style, withTooltip = false, testId }) => {
+    const [isCopied, setIsCopied] = useState(false);
+    const [isTooltipOpen, setIsTooltipOpen] = useState(false);
+
+    const handleCopy = async () => {
+        await navigator.clipboard.writeText(content);
+        setIsCopied(true);
+        setIsTooltipOpen(true);
+        window.dispatchEvent(new Event('resize')); // trigger resize event to update alignment of the tooltip
```
ragi96:
```suggestion
        window.dispatchEvent(new Event('resize'));
```

packages/code-snippet-block/src/components/CopyButton.tsx
```diff
@@ -0,0 +1,73 @@
...
+        debounce(() => {
+            setIsCopied(false);
+            window.dispatchEvent(new Event('resize'));
+        }, COPIED_STATE_DURATION)();
+    };
```
ragi96, line 37: Move this useCopy hook to the shared folder (for the tests etc see in the webapp) 

````

import { useCallback, useEffect, useState } from 'react';

export const useCopy = () => {
    const [status, setStatus] = useState<'success' | 'error' | 'idle'>('idle');

    const copy = useCallback(async (text: string) => {
        setStatus('idle');
        try {
            await navigator.clipboard.writeText(text);
            setStatus('success');
        } catch {
            setStatus('error');
        }
    }, []);

    useEffect(() => {
        let timeout: ReturnType<typeof setTimeout>;

        if (status === 'success' || status === 'error') {
            timeout = setTimeout(() => setStatus('idle'), 2000);
        }

        return () => {
            if (status === 'success' || status === 'error') {
                clearTimeout(timeout);
            }
        };
    }, [status]);

    return { copy, status };
};
````

File: packages/code-snippet-block/src/CodeSnippetBlock.tsx [hunk 4]
```diff
@@ -174,34 +146,19 @@ export const CodeSnippetBlock: FC<BlockProps> = ({ appBridge }) => {
...
-                                            style={getStyle()}
-                                            onClick={handleCopy}
-                                        >
-                                            {isCopied ? <IconCheckMark /> : <IconClipboard />}
-                                        </button>
-                                    </Tooltip.Trigger>
-                                    <Tooltip.Content>{isCopied ? 'Copied' : 'Copy to clipboard'}</Tooltip.Content>
-                                </Tooltip.Root>
+                                <CopyButton
+                                    content={blockSettings.content}
```
ragi96, line 128:
```suggestion
                                    content={blockSettings.content || ''}
```

packages/code-snippet-block/src/components/CopyButton.tsx
```diff
@@ -0,0 +1,73 @@
...
+    /** Renders an icon-only button wrapped in a tooltip instead of an icon with a visible label. */
+    withTooltip?: boolean;
+    testId?: string;
+};
+
+export const CopyButton: FC<CopyButtonProps> = ({ content, className, style, withTooltip = false, testId }) => {
+    const [isCopied, setIsCopied] = useState(false);
+    const [isTooltipOpen, setIsTooltipOpen] = useState(false);
+
+    const handleCopy = async () => {
```
ragi96, line 31: Also use the useCopy hook here

packages/code-snippet-block/src/components/CopyButton.tsx
```diff
@@ -0,0 +1,74 @@
...
+    const [isTooltipOpen, setIsTooltipOpen] = useState(false);
+    const { copy, status } = useCopy();
+    const isCopied = status === 'success';
+
+    useEffect(() => {
+        if (!isTooltipOpen) {
+            return;
+        }
+
+        window.dispatchEvent(new Event('resize'));
```
ragi96, line 28: **[suggestion]** The effect dispatches `window` `resize` whenever the tooltip is considered open and `isCopied` or `isTooltipOpen` changes. The old `handleCopy` only fired resize on copy and 2s later, to realign the tooltip after the icon/label swap. This now also fires on hover (`Tooltip.Root` `onOpenChange(true)`), and because `handleCopy` always calls `setIsTooltipOpen(true)` even when `withTooltip` is false, labeled header/single-line buttons start emitting resize on every later status change too. This block mounts CodeMirror, which typically remasures on window resize.

**Suggestion:** Dispatch resize only when `withTooltip` is true and the copied-state swap actually changes trigger size. Do not set `isTooltipOpen` on the labeled (`withTooltip={false}`) path.

packages/shared/src/hooks/useCopy.ts
```diff
@@ -0,0 +1,33 @@
...
+
+const DEFAULT_RESET_AFTER_MS = 2000;
+
+type CopyStatus = 'success' | 'error' | 'idle';
+
+export const useCopy = (resetAfterMS = DEFAULT_RESET_AFTER_MS) => {
+    const [status, setStatus] = useState<CopyStatus>('idle');
+
+    const copy = useCallback(async (text: string) => {
+        setStatus('idle');
```
ragi96: **[suggestion]** `copy()` sets `status` to `'idle'` before `writeText` so a second success can retrigger the reset effect (`'success'` → `'success'` would bail out). Every consumer that treats `status === 'success'` as the copied UI (CopyButton, SquareBadge, CssValueDisplay) therefore briefly renders the default copy state. The snippet block previously kept `isCopied` true until the 2s timeout, so recopying now flickers Copied → Copy → Copied.

**Suggestion:** Store the timeout id in a ref and reset it inside `copy()` after a successful write, without transitioning through `'idle'`.

packages/code-snippet-block/src/components/CopyButton.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { Tooltip } from '@frontify/fondue/components';
+import { IconCheckMark, IconClipboard } from '@frontify/fondue/icons';
+import { useCopy } from '@frontify/guideline-blocks-shared';
+import { type CSSProperties, type FC, useEffect, useState } from 'react';
+
+type CopyButtonProps = {
+    content: string;
+    className?: string;
+    style?: CSSProperties;
+    /** Renders an icon-only button wrapped in a tooltip instead of an icon with a visible label. */
```
ragi96:
```suggestion
```

packages/code-snippet-block/src/components/CopyButton.tsx
```diff
@@ -18,18 +17,23 @@ export const CopyButton: FC<CopyButtonProps> = ({ content, className, style, wit
...
+    const wasCopiedRef = useRef(false);
 
     useEffect(() => {
-        if (!isTooltipOpen) {
+        if (!withTooltip || (!isCopied && !wasCopiedRef.current)) {
             return;
         }
 
+        wasCopiedRef.current = isCopied;
         window.dispatchEvent(new Event('resize'));
```
ragi96, line 28: Is this still needed

benfrontify:
kontrollieren       ob es das braucht        window.dispatchEvent(new Event('resize'));

packages/code-snippet-block/src/components/CopyButton.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+        if (withTooltip) {
+            setIsTooltipOpen(true);
+        }
+    };
+
+    if (withTooltip) {
+        return (
+            <Tooltip.Root open={isTooltipOpen} onOpenChange={setIsTooltipOpen} enterDelay={0}>
+                <Tooltip.Trigger>
```
ragi96, line 42: **[suggestion]** `Tooltip.Trigger` wraps a native `<button>` without `asChild`. Fondue’s Trigger already renders a button (`asChild` defaults to false), so this produces nested `<button>` elements. Elsewhere in this repo, tooltip+button is done with `asChild` (see `DownloadButton` and `ToolbarButton`). Nested buttons are invalid HTML and can steal or double-handle clicks/focus.

**Suggestion:** Pass `asChild` so the inner `<button>` is the trigger: `<Tooltip.Trigger asChild>`.

packages/code-snippet-block/src/components/CopyButton.tsx
```diff
@@ -0,0 +1,78 @@
...
+                    <button
```
ragi96, line 43: **[suggestion]** The icon-only tooltip button has no accessible name. It only renders `IconCheckMark` / `IconClipboard`, so the control is announced as an unnamed button. The labeled variant is fine because of the visible “Copy” / “Copied” text. Shared copy-like controls in this repo set `aria-label` on the icon button (`DownloadButton`, `ToolbarButton`).

**Suggestion:** Set `aria-label={isCopied ? 'Copied' : 'Copy to clipboard'}` on the tooltip `<button>` (and keep it in sync with the tooltip copy).

packages/code-snippet-block/src/components/CopyButton.tsx
```diff
@@ -0,0 +1,78 @@
...
+    const isCopied = status === 'success';
+    const wasCopiedRef = useRef(false);
+
+    useEffect(() => {
+        if (!withTooltip || (!isCopied && !wasCopiedRef.current)) {
+            return;
+        }
+
+        wasCopiedRef.current = isCopied;
+        window.dispatchEvent(new Event('resize'));
```
ragi96: **[suggestion]** `window.dispatchEvent(new Event('resize'))` is non-obvious: it exists so the Fondue tooltip repositions after the “Copy to clipboard” / “Copied” swap (and CodeMirror typically remasures on window resize). The old inline handler documented that; this effect does not. It also still fires on the success→idle transition when the tooltip is already closed, which is a wasted global resize.

**Suggestion:** Restore a one-line why comment, and only dispatch when the tooltip is actually open (`isTooltipOpen`), so hover-open and closed-idle do not remasure the editor.

packages/shared/src/components/CopyButton/CopyButton.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+type CopyButtonProps = {
+    content: string;
+    className?: string;
+    style?: CSSProperties;
+    withTooltip?: boolean;
+    testId?: string;
+};
+
+export const CopyButton: FC<CopyButtonProps> = ({ content, className, style, withTooltip = false, testId }) => {
```
peter-tudosa, line 19: I think we can move the whole `<CopyButton />` component to `shared/src/components` and use it from there.

packages/code-snippet-block/src/CodeSnippetBlock.tsx
```diff
@@ -118,15 +93,12 @@ export const CodeSnippetBlock: FC<BlockProps> = ({ appBridge }) => {
                             ) : (
                                 <span id={labelId}>{languageNameMap[selectedLanguage]}</span>
                             )}
-                            <button
-                                type="button"
-                                data-test-id="header-copy-button"
+                            <CopyButton
+                                content={blockSettings.content || ''}
```
peter-tudosa: Extract this and use it:
```js
const content = blockSettings.content ?? '';
```
```suggestion
                                content={content}
```

packages/code-snippet-block/src/CodeSnippetBlock.tsx
```diff
@@ -152,34 +124,19 @@ export const CodeSnippetBlock: FC<BlockProps> = ({ appBridge }) => {
...
-                                            style={headerStyle}
-                                            onClick={handleCopy}
-                                        >
-                                            {isCopied ? <IconCheckMark /> : <IconClipboard />}
-                                        </button>
-                                    </Tooltip.Trigger>
-                                    <Tooltip.Content>{isCopied ? 'Copied' : 'Copy to clipboard'}</Tooltip.Content>
-                                </Tooltip.Root>
+                                <CopyButton
+                                    content={blockSettings.content}
```
peter-tudosa: Re-use var:
```suggestion
                                    content={content}
```

packages/code-snippet-block/src/CodeSnippetBlock.tsx
```diff
@@ -152,34 +124,19 @@ export const CodeSnippetBlock: FC<BlockProps> = ({ appBridge }) => {
...
+                                    testId="copy-button"
+                                    className="tw-p-2 tw-rounded-md"
+                                    style={headerStyle}
+                                    withTooltip
+                                />
                             ) : (
-                                <button
-                                    type="button"
+                                <CopyButton
+                                    content={blockSettings.content || ''}
```
peter-tudosa: Reuse var:


```suggestion
                                    content={content}
```

packages/code-snippet-block/src/CodeSnippetBlock.tsx
```diff
@@ -152,34 +124,19 @@ export const CodeSnippetBlock: FC<BlockProps> = ({ appBridge }) => {
                     {!withHeading && (
                         <div className="tw-absolute tw-p-1 tw-dark tw-top-0 tw-right-0 tw-hidden group-hover/copy:tw-block">
                             {blockSettings.content && (blockSettings.content.match(/\n/g) || []).length > 1 ? (
```
peter-tudosa: Reuse `content` var here as well and extract:

```
let isMultiline = (content.match(/\n/g) || []).length > 1
```

peter-tudosa (approved):

PR #1467, opened 2026-09-10
chore: release audio block

fulopdaniel:
Round 2.

Previously the "Clean" job failed, because we don't have a GitHub App for that. This approach uses a PAT to commit directly to main.


peter-tudosa (approved):
We should find a way that doesn't involve our users to do this, in the future.

